PR #60, opened 2025-11-10
ZA | 25-SDC-July | Luke Manyamazi | Sprint 2 | Imrove With PreComputing

Luke-Manyamazi:
## Learners, PR Template

Self checklist

## Changelist

- Optimized find_longest_common_prefix using sorting to reduce comparisons from O(n²) to O(n log n)
- Enhanced count_letters with fixed-size arrays for O(1) lookups instead of dictionary operations
- Both functions now handle edge cases efficiently with minimal memory overhead

## Questions

1. Why does the test_long_string test hang when included, but the other 4 tests pass when it's commented out?
2. Is the issue with generating a 10 million character string or with the count_letters function performance?
3. Should we optimize the string generation logic or add a timeout to prevent hanging?
4. Could the problem be with the still_to_include set logic creating an infinite loop?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .DS_Store
 node_modules
-.venv
+venv
 __pycache__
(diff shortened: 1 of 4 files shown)


OracPrime (approved):
Nice code

illicitonion:

PR #72, opened 2026-01-13
West Midlands | 25-SDC-July | Gabriel Deng | Sprint 2 | Improve with precomputing

gai93003:

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,18 +7,26 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if not strings or len(strings) == 1:
+        return ""
+    
     longest = ""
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Can you explain why your solutions are better?

gai93003:
Yes of course, I can explain.

For count_letters, assignment, The first thing that my code improves is the precomputation of the lowercase_letters in a set. This sets time complexity to O(n) and instead of using (if letter.lower() not in s), I am using set lookup for our precomputed set making the complexity to be O(1) per lookup. 
~ This  improves the process especially if we are working with long strings with alot of uppercase letters.

The overall complexity is O(n) instead of O(n^2) as it was for the original code.

gai93003:
For find_longest_common_prefix

Instead of comparing every pair of strings like in original code which gives a complexity of O(n^2 * m), I build the prefix dictionary once and then scan through it to find the longest prefix appearing alteast 2 times. This gives ~ O(1) for lookups.

The overall complexity for this solution is O(n * m^2) which is good for many short strings

cjyuan:
Analysis looks good. Well done.

gai93003:
Thank you.

illicitonion:

PR #77, opened 2026-01-17
London | 25-SDC-July | Fatma Arslantas | Sprint 2 | Improve code with precomputing

AFatmaa:
## Learners, PR Template

Self checklist

## Changelist

I optimised `common_prefix` and `count_letters` using **Precomputing** techniques.

*   **Common Prefix:**
    *   **Old:** Compared every string with every other string.
    *   **New:** I **sorted** the list first. Similar strings became neighbours. I only compared neighbours.
*   **Count Letters:**
    *   **Old:** Scanned the huge string repeatedly for every uppercase letter.
    *   **New:** I created a **Set** of all characters first. Checking if a lowercase letter exists is now instant.

## Questions

I don’t have any questions, thank you!

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,12 +7,24 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    # If list is empty or has 1 item, no pairs exist.
+    if len(strings) < 2:
+        return ""
+    
+    strings.sort()
+
     longest = ""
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
How do you know your solutions are better?

AFatmaa:
Hi @cjyuan,

Thank you for the feedback!

For your question, I thought my solution is better for two reasons:

First, I analysed the old version that used nested loops (loops inside loops). I tried to improve this by using Sets and Sorting so the code doesn't need to check the same things repeatedly. And second, I ran the tests locally, and the heavy tests finished in 8 seconds. The old code would take much longer or freeze with that amount of data.

Do you think this logic is right?

cjyuan:
You were in the right direction. A more convincing proof would be to compare their complexity using the big-O notations.

illicitonion:

PR #85, opened 2026-01-30
London | 25-SDC-July | Rihanna Poursoltani | Sprint 1 | Analyse and Refactor Functions

RihannaP:
## Learners, PR Template

Self checklist

## Changelist
I could understand code that most of them had complexity with nested loops. I refactored them by using Set and one loops to help complexity in both JS and PY.

## Questions
No question

### DIFF
--- a/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
+++ b/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
@@ -1,14 +1,15 @@
 /**
  * Finds common items between two arrays.
  *
- * Time Complexity:
- * Space Complexity:
- * Optimal Time Complexity:
+ * Time Complexity: O(n + m)
+ * Space Complexity: O(n)
+ * Optimal Time Complexity: O(n)
  *
  * @param {Array} firstArray - First array to compare
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
-export const findCommonItems = (firstArray, secondArray) => [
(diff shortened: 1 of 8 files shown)


Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -1,14 +1,15 @@
...
  * @param {Array} firstArray - First array to compare
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
-export const findCommonItems = (firstArray, secondArray) => [
-  ...new Set(firstArray.filter((item) => secondArray.includes(item))),
-];
+export const findCommonItems = (firstArray, secondArray) => {
+  const lookup = new Set(secondArray);
+  return [...new Set(firstArray.filter(item => lookup.has(item)))];
```
LonMcGregor, line 14: Given you are using sets, are there any built-in methods that might help make this code even simpler?

LonMcGregor (commented):
Good work on these, you have found some good optimisations.

There are a few times you identify nested loops as o(n). If you have a nested loop, with the same array looping each time, are you sure it is going to be o(n)?

See also my other comment, there may be better ways to use sets in javascript.

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
RihannaP: Good point, since I am already using sets, I can simplify this by converting both arrays to Sets and using Set.prototype.intersection() to directly compute the common elements.

illicitonion:

PR #88, opened 2026-01-30
London | 25-SDC-July | Rihanna Poursoltani | Sprint 2 | Improve code with caches

RihannaP:
## Learners, PR Template

Self checklist

## Changelist
I tried to improve the performance by :
- adding memo dictionary
- adding cache using tuple keys and solve problem with math (not loop)

## Questions
No question

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,12 +1,13 @@
 from typing import List
-
-
+cache = {}
 def ways_to_make_change(total: int) -> int:
     """
     Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
 
     For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
     """
+  
+    cache.clear()
     return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
 
 
@@ -14,9 +15,20 @@ def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
     """
     Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
     """
+    key = (total, tuple(coins))
+
+    if key in cache:
+        return cache[key]
+    
     if total == 0 or len(coins) == 0:
         return 0
-
+    
+    if len(coins) == 1:
+        if total % coins[0] == 0:
+            return 1
+        else:
+            return 0
+        
     ways = 0
     for coin_index in range(len(coins)):
         coin = coins[coin_index]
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py

From line 41, we know `coins` can only be one of the following 8 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[1]
```

     - There are only a small number of different subarrays. We can map each subarray to a unique integer. 

Can you implement this?

RihannaP: Yes, since coins is always one of 8 suffixes of the same base list, we can pre-create those 8 suffix arrays once and then pass around a unique integer id . So it avoid repeated sub-array creation and use a smaller cache cache key

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,57 @@
-from typing import List
 
+from typing import List, Dict, Tuple
 
-def ways_to_make_change(total: int) -> int:
-    """
-    Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
+# Pre-create only 8 possible "coins" arrays
+_BASE_COINS: List[int] = [200, 100, 50, 20, 10, 5, 2, 1]
+COIN_SUFFIXES: List[Tuple[int, ...]] = [tuple(_BASE_COINS[i:]) for i in range(len(_BASE_COINS))]
```
cjyuan: We could also just use the starting index to identify the sub array without creating the sub array.

0 => whole array
1 => the sub array starting at index 1
...

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,57 @@
...
+
+    # Keep behavior close to original, but fix the standard base case:
+    if total == 0:
+        return 1
+    if len(coins) == 0:
         return 0
 
+    if len(coins) == 1:
+        cache[key] = 1 if (total % coins[0] == 0) else 0
+        return cache[key]
```
cjyuan: Note: Even though tuple construction and dictionary lookup (lines 21 and 22) are O(1) operations, 
they are still relatively costly than some simple numerical computation and comparisons (lines 28, 30, 33, 34).

This is unrelated to cache, but if you swap the code on lines 28-33 with the code on lines 21-23,
you can probably notice some slight improvement in performance.

---
Good job in figuring out a quicker way to compute number of ways when there is only one coin left to consider.

cjyuan (commented):
Code is good.

Thanks for updating the label for me.

Sprint-2/improve_with_caches/making_change/making_change.py
RihannaP: Yes, thank you for mentioning that

Sprint-2/improve_with_caches/making_change/making_change.py
RihannaP: Agreed. Checking that before doing cache lookup avoids unnecessary tuple creation and dictionary access

illicitonion:

PR #89, opened 2026-01-30
London | 25-SDC-July | Rihanna Poursoltani | Sprint 2 | Improve code with precomputing

RihannaP:
## Learners, PR Template

Self checklist

## Changelist
For optimising I tried to
- Store the list and the compare only with neighbours
- I added a set of all char and then check  if a lowercase letter exists

## Questions
No question

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,12 +7,21 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if len(strings) < 2:
+        return ""
+    
+    strings.sort()
     longest = ""
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
-            common = find_common_prefix(string, other_string)
-            if len(common) > len(longest):
-                longest = common
+
+    for i in range(len(strings) - 1):
+        first_string = strings[i]
+        second_string = strings[i + 1]
+        
+        common = find_common_prefix(first_string, second_string)
+        
+        if len(common) > len(longest):
+            longest = common
+            
     return longest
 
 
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Code looks good. Can you explain using big-O notation why your approach is better?

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
```diff
@@ -7,12 +7,21 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if len(strings) < 2:
+        return ""
+    
+    strings.sort()
```
cjyuan: `strings.sort()` can modify the original list passed into the function by the caller.
How can you prevent this side effect?

RihannaP: Good point!
To avoid this side effect, we can sort a copy of the list instead .

RihannaP:
> Code looks good. Can you explain using big-O notation why your approach is better?

Using a set makes membership checks O(1) on average. Creating the set takes O(n) and we iterate over at most n unique characters, so the overall complexity is O(n).

cjyuan:
> > Code looks good. Can you explain using big-O notation why your approach is better?
> 
> Using a set makes membership checks O(1) on average. Creating the set takes O(n) and we iterate over at most n unique characters, so the overall complexity is O(n).

Can you also do the same for `common_prefix.py`?

RihannaP:
> Can you also do the same for `common_prefix.py`?

Sorting the strings takes O(n log n). After sorting, we only compare adjacent strings, and each prefix comparison takes O(m) ( m is the length of the shorter string). This results in O(n·m) for the comparisons.

Overall complexity is O(n log n + n·m), which is better than comparing every pair of strings (O(n²·m)).

cjyuan:
If we are factoring in the length of the strings, `m`, then the complexity of sorting won't just be `O(n log n)`.
`n log n` measures only the number of comparisons needed.

RihannaP:
> If we are factoring in the length of the strings, `m`, then the complexity of sorting won't just be `O(n log n)`. `n log n` measures only the number of comparisons needed.

You’re right. so sorting and comparison cost is O(m · n log n)
Adjust scan O(n · m) so overall:
O(n log n · m + n · m)= O(n log n · m)

cjyuan:
Yes. O(m · n log n)

O(n log n · m) may be misinterpreted as O(n log (n · m)).

illicitonion:

PR #90, opened 2026-01-31
London|25-SDC-November|Donara Blanc |Sprint 2| Lru-cache

donarbl:

## Change List

I separated DoublyLinkedList from LruCache (SRP) to follos the recommendation of Solid design principle

### DIFF
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -1,91 +1,89 @@
 class _Node:
-    def __init__(self, key, value):
-        self.key = key
-        self.value = value
+    def __init__(self, value):
+        self.value = value  # stores as (key, value) tuple
         self.next = None
         self.previous = None
 
 
 class DoublyLinkedList:
+    """responsible only for managing a doubly linked list.
+    stores values as (key, value) tuples but knows nothing about cachin
+    """
     def __init__(self):
-        self.head = None
-        self.tail = None
+        self.head = None  # most recently used end
+        self.tail = None  # least recently used end
 
-    def _add_to_head(self, node):
-        """Put node at the front (most recently used)."""
-        node.previous = None
+    def add_to_head(self, value):
+        """sreates a new node with value, add to head, returns the node."""
+        node = _Node(value)
         node.next = self.head
-
+        node.previous = None
         if self.head is not None:
             self.head.previous = node
         self.head = node
-
         if self.tail is None:
-            # First node in the list
             self.tail = node
+        return node
 
-    def _remove_node(self, node):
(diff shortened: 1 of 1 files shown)


cjyuan (commented):

Alternatively, `OrderedDict` can be used directly within `LruCache` to maintain order.

donarbl:
Thank you for suggestion and it is done

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,91 @@
...
+        self.next = None
+        self.previous = None
+
+
+class DoublyLinkedList:
+    def __init__(self):
+        self.head = None
+        self.tail = None
+
+    def _add_to_head(self, node):
```
cjyuan: What's your reason to to make the methods "private" (by naming convention)?

This could be just a normal linked list.
You could even import the LinkedList you implemented in the other exercise and use it here, and represent the value of `data` as `(key, value)`.

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,91 @@
...
+        if node is not None:
+            # update existing
+            node.value = value
+            self._move_to_head(node)
+            return
+
+        # new key
+        if len(self.map) >= self.limit:
+            # evict least recently used (tail)
+            lru = self.tail
```
cjyuan: Does `LruCache` has a `tail` property?

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,91 @@
...
+            if lru is not None:
+                self._remove_node(lru)
+                del self.map[lru.key]
+
+        new_node = _Node(key, value)
+        self._add_to_head(new_node)
```
cjyuan, line 88: It would make the LinkedList easier to use if `add_to_head()` is designed in such a way that the caller can push a new node to the front as:
```
    # Just need to add the data; does not need to concern node creation logic
    new_node = self.list.add_to_head((key, value)); # Add key+value as a tuple.
```

cjyuan: It seems some of the properties and methods accessed here belong to `LinkedList` (not to `LruCache`).

cjyuan (commented):
I don't think the current implementation can pass the tests in `lru_cache_test.py`.

Sprint-2/implement_lru_cache/lru_cache.py
donarbl: I have separate the DoublyLinkedList and LruCache
They don't overlap. LruCache never touches node pointers directly,it always goes through self.list.

PR #94, opened 2026-02-09
London | 25-SDC-NOV | Jesus del Moral | Sprint 1 | Analyse and Refactor Functions

delmorallopez:


Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
```diff
@@ -9,21 +9,23 @@
...
- * Time Complexity:
- * Space Complexity:
- * Optimal Time Complexity:
+ * Time Complexity: First and Second loop both iterate through all n elements to calculate sum and product,
+ * O(n) + O(n) = O(2n) = O(n), Two sequential passes through the array
+ * Space Complexity: O(1), Only a constant amount of space is used for the sum and product variables, regardless of input size
+ * Optimal Time Complexity: O(n) - We must examine each element at least once to calculate the sum and product, so O(n) is the best we can achieve for this problem.
  *
+ * we can reduce the constant factor by eliminating the redundant second loop
+ * Instead of two passes (2n operations), we can do both calculations in a single pass (n operations)
```
cjyuan, line 18: What do you mean by "two passes == `2n` operations" and "1 pass == `n` operation"?
What are these operations? Does that mean one pass is guaranteed to take half as much time to complete?

The code still needs to perform `sum += num` and `product *= num` whether we use one loop or two loops.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -9,13 +9,11 @@ def find_common_items(
...
+    Optimal time complexity: O(n + m) - We must examine each element of both sequences at least once to find common items
     """
-    common_items: List[ItemType] = []
-    for i in first_sequence:
-        for j in second_sequence:
-            if i == j and i not in common_items:
-                common_items.append(i)
-    return common_items
+
+    return list(set(first_sequence) & set(second_sequence))
```
cjyuan, line 17: Clean solution!

cjyuan (commented):
Very detailed analysis.

Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
delmorallopez: When I said “2n vs n operations”, I meant loop iterations. However, both implementations still perform n additions and n multiplications. The single-loop version reduces loop overhead and improves cache efficiency, but asymptotically both are O(n). One pass is not guaranteed to take half the time, because Big-O ignores constant factors and the arithmetic work remains the same.

cjyuan: You have a good understanding and the performance is indeed improved when two loops are reduced to one.
However, the improvement is not really "doubled", so in complexity analysis we don't focus on the constant factor.

illicitonion:

PR #95, opened 2026-02-10
Glasgow | 25-SDC-Nov | Nataliia Volkova | Sprint 1 | Analyse and Refactor Functions

Nataliia74:
## Learners, PR Template

Self checklist

I used to solve these tasks with a hash set(based on hash table logic, which is very fast in terms of time complexity) for constant-time lookups and processes each element once. In addition, for JS logic new Set also keeps ordering along with deduplication.

### DIFF
--- a/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
+++ b/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
@@ -14,10 +14,11 @@
 // ];
 
 export const findCommonItems = (firstArray, secondArray) => {
-  let firstArr = new Set(firstArray);
-  let secondArr = new Set(secondArray);
+  const firstArr = new Set(firstArray);
+  const secondArr = new Set(secondArray);
 
-  return [...firstArr].filter((arr) => secondArr.has(arr));
+  // return [...firstArr].filter((arr) => secondArr.has(arr));
+  return [...firstArr.intersection(secondArr)];
 };
 
 //  * https://www.w3schools.com/js/js_set_methods.asp#mark_set_new
--- a/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
+++ b/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
@@ -9,6 +9,18 @@
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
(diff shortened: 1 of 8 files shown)


Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -9,6 +9,18 @@
...
-export const findCommonItems = (firstArray, secondArray) => [
-  ...new Set(firstArray.filter((item) => secondArray.includes(item))),
-];
+// export const findCommonItems = (firstArray, secondArray) => [
+//   ...new Set(firstArray.filter((item) => secondArray.includes(item))),
+// ];
+
+export const findCommonItems = (firstArray, secondArray) => {
+  let firstArr = new Set(firstArray);
+  let secondArr = new Set(secondArray);
```
cjyuan: For variables that do not need to be reassigned a value, better practice is to declare them using `const`.

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -9,6 +9,18 @@
...
+
+  return [...firstArr].filter((arr) => secondArr.has(arr));
```
cjyuan: Note: Set has a built-in method for finding common elements between two sets.

cjyuan (commented):
I think in this exercise,
"Time Complexity" is supposed to be the complexity of the original code, and 
"Optimal Time Complexity" is the complexity of the refactored code. 

Can you also describe the complexity of the original code?

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
Nataliia74: Hello CJ Yuan, thank you for spotting. I adjusted it.

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
Nataliia74: I found it. Thanks.

Nataliia74:
> I think in this exercise, "Time Complexity" is supposed to be the complexity of the original code, and "Optimal Time Complexity" is the complexity of the refactored code.
> 
> Can you also describe the complexity of the original code?

It will be: O(m*n), because includes is nested inside filter.

PR #101, opened 2026-02-15
Sheffield | 25-SDC-Nov | Sheida Shabankari | Sprint 2 |Implement linked list in python

sheida-shab:

This PR implements a doubly linked list in Python with O(1) time complexity for all required operations.

Added a Node class to represent elements with value, next, and previous references.

Implemented push_head to insert elements at the head and return a node handle for constant-time removal.

Implemented pop_tail to remove and return the value of the tail node.

Implemented remove to delete a node from the list using its handle, without traversing the list.

Properly maintains head and tail references and updates links for all edge cases (empty list, single node, head, tail, and middle nodes).

All provided tests pass successfully.

### DIFF
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -1,0 +1,24 @@
+class Node:
+    """
+    Represents a node in a doubly linked list.
+    Each node holds a value and references to the previous and next nodes.
+    """
+    def __init__(self,value):
+        self.value=value
+        self.next=None
+        self.previous=None
+
+class LinkedList:
+    """
+    A doubly linked list supporting push_head, pop_tail, and remove operations.
+    Maintains references to both head and tail nodes for O(1) operations.
+    """
+    def __init__(self):
+        self.head=None
+        self.tail=None        
+    def push_head(self,value) -> Node:
+        pass
+    def pop_tail(self) -> any:
+        pass
+    def remove(self,node) -> None:
+        pass
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -32,13 +32,14 @@ class LinkedList:
         if self.head is None:
             return None
         elif self.head==self.tail:
-            node_value=self.head.value
-            self.head=None
-            self.tail=None
+            node_to_remove = self.head
+            node_value = node_to_remove.value
+            self.remove(node_to_remove)
+            
         else:
-            node_value=self.tail.value
-            previous_node=self.tail.previous   
-            self.tail=previous_node 
+           node_to_remove = self.tail
+           node_value = node_to_remove.value
+           self.remove(node_to_remove)
         return node_value    
     
     def remove(self,node) -> None:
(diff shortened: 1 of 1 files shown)


Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,69 @@
...
+        if self.head is None:
+            return None
+        elif self.head==self.tail:
+            node_value=self.head.value
+            self.head=None
+            self.tail=None
+        else:
+            node_value=self.tail.value
+            previous_node=self.tail.previous   
+            self.tail=previous_node
```
cjyuan: Could also consider calling `remove()` to remove the tail -- less code to maintain.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,69 @@
...
+            self.head.previous=None
+        else:
+            #Node is in the middle of list
+            previous_node=node_to_remove.previous        
+            next_node=node_to_remove.next
+            previous_node.next=next_node
+            next_node.previous=previous_node
+        
+        node_to_remove.next=None
+        node_to_remove.previous=None
```
cjyuan: Why not also clearing the reference of the removed tail in `pop_tail()`?

Sprint-2/implement_linked_list/linked_list.py
sheida-shab: Did it.

Sprint-2/implement_linked_list/linked_list.py
sheida-shab: Did it.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,69 @@
...
+            return None
+        elif self.head==self.tail:
+            node_to_remove = self.head
+            node_value = node_to_remove.value
+            self.remove(node_to_remove)
+            
+        else:
+           node_to_remove = self.tail
+           node_value = node_to_remove.value
+           self.remove(node_to_remove)
```
cjyuan: Wouldn't this work the same as the code on line 34-42?
```
  node_value = self.tail.value
  self.remove(self.tail)
```

cjyuan (commented):
Changes look good.

cjyuan:
Just noticed you removed these two lines from `removed()`:
```
        node_to_remove.next=None
        node_to_remove.previous=None     
```

Why?

sheida-shab:
You are right.
After the refactor, pop_tail() no longer removes nodes directly.
That responsibility was moved to remove(), so clearing the node references should also stay in remove().
I understand this now and will keep the code there.

Sprint-2/implement_linked_list/linked_list.py
sheida-shab: I understand the point now.
remove() handles both single-node and multiple-node cases,
so I simplified pop_tail() to just call remove(self.tail).
This makes the code shorter and cleaner.

cjyuan:
I could not find the code for clearing references in `remove()`.

sheida-shab:
You’re right, that was an oversight on my part.
I’ve added the reference cleanup back into remove().

cjyuan:
All good now. Well done.

illicitonion:

PR #105, opened 2026-02-16
Sheffield | 25-SDC-Nov | Sheida Shabankari | Sprint 2 | Implement LRU cache in python

sheida-shab:

PR Summary :
Implement LRU Cache with O(1) get/set operations

Description:
This PR adds a fully functional LRU (Least Recently Used) cache in Python. The implementation includes:

Node class: Represents a key-value pair with pointers to previous and next nodes.

Doubly linked list: Maintains the order of nodes for LRU eviction in O(1) time.

push_head: Moves or adds a node to the head (most recently used).

pop_tail: Removes and returns the tail node (least recently used).

remove: Removes a specific node from the list.

LruCache class: Provides the cache interface.

get(key): Retrieves a value by key and marks the node as most recently used.

set(key, value): Adds or updates a key-value pair; evicts the least recently used node if the cache exceeds its limit.

Handles invalid cache limits by raising ValueError.

Each operation (get and set) has O(1) worst-case time complexity.

Testing:

Unit tests cover cache limit enforcement, eviction order, and get/set functionality.


Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self,limit):
+        """
+        Initialize the LRU Cache.
+        Args: limit (int): Maximum number of items the cache can hold.
+        """
+        if limit<=0:
+            raise ValueError("limit must be greater than zero")
+        self.limit=limit
+        self.map={}
+        self.List=LinkedList()
```
cjyuan: Why start the name the property `List` with an uppercase `L`?

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,150 @@
...
+    Maintains references to both head and tail nodes for O(1) operations.
+    """
+    def __init__(self):
+        self.head=None
+        self.tail=None    
+            
+    def push_head(self,node=None,key=None,value=None) -> Node:
+        """
+        Adds a node to the head of the list.
+        If 'node' is provided, it is moved to the head.
```
cjyuan: Moving a node to the front does not seem like a normal behavior expected from the "push head" operation.

You could consider implementing a helper function or just have the caller explicitly 
1. remove the node to be moved, and
2. add that node to the front of the list

cjyuan (commented):
- Why not simply import or subclass the Double Linked List you implemented in the other exercise (to practice code reuse)?
 
- Alternatively, you can explore using `OrderedDict` within `LruCache` to maintain order.

Could you improve your `LruCache` implementation using one of these approaches?

Sprint-2/implement_lru_cache/lru_cache.py
sheida-shab: I wasn’t paying attention to Python’s naming conventions for properties. I realized that property names should be lowercase, so I’ve renamed List to list to follow the standard convention.

Sprint-2/implement_lru_cache/lru_cache.py
sheida-shab: I did it.

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,58 @@
...
+        else:
+            if self.count>=self.limit:
+                old_key,old_value=self.list.pop_tail()
+                del self.map[old_key]
+                self.count -=1
+
+            new_node=self.list.push_head(key=key,value=value)
+            self.count +=1
+            self.map[key]=new_node
+
```
cjyuan, line 55: It is possible to use the `LinkedList` you implemented in the "Linked List" exercise (without modification).
We just have to store the key-value pair as a tuple (or any structured data type).

The drawback is, we can't reuse the node when we want to move it to the front.
```python
        if key in self.map:
            node=self.map[key] 
            self.list.remove(node) 
            self.map[key] = self.list.push_head((key, value))
        else:
            if self.count>=self.limit:
                old_key, _ = self.list.pop_tail()
                del self.map[old_key]
                self.count -=1

            self.map[key] = self.list.push_head((key, value))
            self.count +=1
```

cjyuan (commented):
Code works.

Could consider using some tool to auto format the code to add spacing around operators consistently (which improves readability).

Sprint-2/implement_lru_cache/lru_cache.py
sheida-shab: Thanks for your useful feedbacks.

illicitonion:

PR #106, opened 2026-02-16
Glasgow | 25-SDC-Nov | Nataliia Volkova | Sprint 2 | improve_with_caches

Nataliia74:
## Learners, PR Template

Self checklist

## Changelist
Improve code with caches 11

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,5 +1,6 @@
 from typing import List
 
+cache = {}
 
 def ways_to_make_change(total: int) -> int:
     """
@@ -8,14 +9,25 @@ def ways_to_make_change(total: int) -> int:
     For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
     """
     return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
+    
 
 
 def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+    key = (total, tuple(coins))
+    if key in cache:
+        return cache[key]
     """
     Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
     """
     if total == 0 or len(coins) == 0:
-        return 0
+        result = 0
+        cache[key] = result
+        return result
+
+    if total == 1 or len(coins) == 1:
+        result = 1
+        cache[key] = result
+        return result
 
     ways = 0
     for coin_index in range(len(coins)):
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -8,14 +9,25 @@ def ways_to_make_change(total: int) -> int:
...
     if total == 0 or len(coins) == 0:
-        return 0
+        result = 0
+        cache[key] = result
+        return result
+
+    if total == 1 or len(coins) == 1:
+        result = 1
+        cache[key] = result
+        return result
```
cjyuan: When `total == 1 or len(coins) == 1` is true, the result may not be 1 if the last coin is not 1 (e.g. the initial list of coins given is different)

On the other hand, when `len(coins) == 1`, there is simple way to determine if the current total (any value) can be made by the coin value.

Sprint-2/improve_with_caches/making_change/making_change.py

From line 41, we know `coins` can only be one of the following 9 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[1]
[]
```

Sprint-2/improve_with_caches/making_change/making_change.py
Nataliia74: Hello CJ Yuan, I understood that I can have 2 possible results, not just 1. So, the best founded solution is not to use recursion, but simple math to check if the total is divisible by that left type coin.

cjyuan: What do you mean by 2 possible results? 0 or 1?

Yes, this approach:
>  simple math to check if the total is divisible by that left type coin

Sprint-2/improve_with_caches/making_change/making_change.py
Nataliia74: Eliminated slicing, which is costly and creates a new array every recursion, and also many duplicates; the tuple conversion was costly as well. Changed by exchanging the tuple at the index an array.

Sprint-2/improve_with_caches/making_change/making_change.py
Nataliia74: Hello CJ Yuan, I mean 0 or 1. In this case, a better approach is to use math.

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,38 @@
 from typing import List
 
+cache = {}
+coins = [1, 2, 5, 10, 20, 50, 100, 200]
```
cjyuan, line 4: Should you need to avoid introducing variables in the global scope, you can consider:
- Declare them in `ways_to_make_change()`
- Define `ways_to_make_change_helper()` as an inner function of `ways_to_make_change()`. Inner function can access the variables in the outer scope.

cjyuan (commented):
Changes look good.

illicitonion:

PR #110, opened 2026-02-16
Remove peer review instructions

illicitonion:
This is currently confusing/unclear as to whether volunteers should be reviewing PRs or leaving them to trainees.

We may want to re-add peer review, but if we do, we should make sure it's clear what the volunteer expectations are.


OracPrime (approved):
Looks good to me

PR #112, opened 2026-02-17
Sheffield | 25-SDC-Nov | Sheida Shabankari | Sprint 2 |improve with caches

sheida-shab:

PR summary :

Add manual caching to speed up recursive Fibonacci and ways_to_make_change functions.

- Use dictionaries to store results of already computed subproblems.
- In Fibonacci, store `fibonacci(n)` values to avoid recalculating them.
- In ways_to_make_change, use a key `(total, tuple(coins))` to remember results for each combination of total and remaining coins.
- All recursive calls share the same cache to reuse previous results, without using `@cache`.
- These changes reduce repeated calculations and make the functions much faster for large inputs.

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -10,13 +10,18 @@ def ways_to_make_change(total: int) -> int:
     return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+def ways_to_make_change_helper(total: int, coins: List[int],cache=None) -> int:
     """
     Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
     """
     if total == 0 or len(coins) == 0:
         return 0
-
+    if cache is None:
+        cache={}
+    key=(total,tuple(coins))    
+    if key in cache:
+        return cache[key]
+    
     ways = 0
     for coin_index in range(len(coins)):
         coin = coins[coin_index]
@@ -26,7 +31,8 @@ def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
             if total_from_coins == total:
                 ways += 1
             else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
+                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:],cache=cache)
                 ways += intermediate
             count_of_coin += 1
+    cache[key]=ways        
     return ways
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -10,13 +10,18 @@ def ways_to_make_change(total: int) -> int:
     return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+def ways_to_make_change_helper(total: int, coins: List[int],cache=None) -> int:
```
cjyuan: Why not use `{}` as the default value for cache?

Sprint-2/improve_with_caches/making_change/making_change.py

From line 34, we know `coins` can only be one of the following 9 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[1]
```

Sprint-2/improve_with_caches/making_change/making_change.py
sheida-shab: If I use {} as the default value, the cache is shared between all function calls .
Not only between recursive calls, but between every call to the function in the whole program.

This happens because default arguments in Python are created once, when the function is defined.

If I use None instead, then every time ways_to_make_change is called,
a new cache dictionary is created.

At the same time, all recursive calls share that same cache,
which is exactly what we want for memoisation .

cjyuan: Thanks for clarification. I didn't know that. 

Here are two alternatives,
1. Create the cache in `ways_to_make_change()`  and then pass it to `ways_to_make_change_helper()`, or 

2. Define `ways_to_make_change_helper()` as an inner function of `ways_to_make_change()`, and then have 
the inner function share the cache declared in `ways_to_make_change()`.

Sprint-2/improve_with_caches/making_change/making_change.py
sheida-shab: Refactor memoisation to avoid repeated sub-array creation

Precompute all possible coin sub-arrays once and identify them by index.
Use (total, subarray_id) as the cache key instead of (total, tuple(coins)),
so recursive calls reuse the same sub-arrays and avoid unnecessary list
and tuple allocations.

Sprint-2/improve_with_caches/making_change/making_change.py
sheida-shab: Thanks for the suggestions!

I’ve updated the code again based on your guidance. I now create the cache in
ways_to_make_change() and pass it into the helper function, so the cache is scoped
to a single top-level call but shared across all recursive calls.

cjyuan:
Looks good. Well done.

illicitonion:

PR #113, opened 2026-02-18
Sheffield | 25-SDC-Nov | Sheida Shabankari | Sprint 2 |Improve code with precomputing

sheida-shab:

PR summary : 

* Improved find_longest_common_prefix using two techniques:
1. Precomputing prefix hashes for each string to avoid repeated character-by-character comparisons.
2. Sorting the strings so that only adjacent strings need to be compared, which greatly reduces the number of comparisons for large lists.
These changes make the function much faster while keeping all existing tests passing, including the large list test.

* Improved count_letters by precomputing lowercase letters for faster lookup. Only uppercase letters without a lowercase version are counted.  All tests, including long strings, pass.

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,17 +7,23 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    # If fewer than 2 strings, no common prefix is possible
+    if len(strings)<2:
+        return ""
+    # Sort strings to ensure only adjacent strings need comparison
+    strings.sort()
+    
     longest = ""
 
     # Precompute prefix hashes for each string to speed up comparisons
     prefix_map={s:[hash(s[:i+1]) for i in range(len(s))] for s in strings}
 
-    # Compare all pairs of strings using precomputed hashes
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
-            common = find_common_prefix(string, other_string,prefix_map)
-            if len(common) > len(longest):
-                longest = common
+    # Compare only adjacent strings in the sorted list
+    for i in range(len(strings)-1):
+        common=find_common_prefix(strings[i],strings[i+1],prefix_map)
+        if len(common)>len(longest):
+            longest=common
+
     return longest
 
 
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -8,17 +8,31 @@ def find_longest_common_prefix(strings: List[str]):
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
     longest = ""
+
+    # Precompute prefix hashes for each string to speed up comparisons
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
```diff
@@ -7,18 +7,38 @@ def find_longest_common_prefix(strings: List[str]):
...
+    
     longest = ""
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
-            common = find_common_prefix(string, other_string)
-            if len(common) > len(longest):
-                longest = common
+
+    # Precompute prefix hashes for each string to speed up comparisons
+    prefix_map={s:[hash(s[:i+1]) for i in range(len(s))] for s in strings}
```
cjyuan: How do these "prefix hashes" speed up comparison?

cjyuan (commented):
Can you use complexity to explain why your implementation is an improvement?

sheida-shab:
Hi CJ,Thanks for your feedback.
I improved count_letters by reducing time complexity from O(n²) to O(n).

Previously, for each character in the string (O(n)), the code performed a membership check on the entire string (O(n)), resulting in  O(n²)  complexity.

By precomputing a set of all lowercase characters once (O(n)) and using O(1) set lookups inside the loop, the overall complexity is reduced to O(n) .

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
sheida-shab: At first, I thought the exercise needed me to precompute and keep some data for each string. So I used prefix hashes, thinking this would avoid comparing letters one by one and make the function much faster.
But after testing, I saw that using hashes did not make it much faster. Sorting the strings and comparing only neighbors worked better, was simpler, and gave the speed improvement needed.

cjyuan: Note: With the way you used prefix hashes, the complexity actually became higher (when the length of each string is large).

cjyuan:
Can you also do a complexity analysis for `find_longest_common_prefix`?

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
sheida-shab: That makes sense. In my case, prefix hashing introduced additional preprocessing and memory costs without improving the asymptotic complexity. Once I removed it and only compared adjacent strings after sorting, the solution became simpler and faster, especially for longer inputs.

sheida-shab:
The time complexity of find_longest_common_prefix  now is O(n log n + n · m), where n is the number of strings and m is the average length of a string.
Sorting the strings takes O(n log n), and then comparing only adjacent strings takes O(n · m) in the worst case.

cjyuan:
How does the program determine the order between two strings? Is the performance affected by string length when we compare two strings? If we take into account `m`, then the sorting complexity is no longer just O(nlogn).

sheida-shab:
Thanks for your feedback!
The ordering between two strings during sorting is determined lexicographically in Python, by comparing characters one by one until a difference is found or one string ends.
Comparing two strings therefore takes O(m) time in the worst case, where m is the string length.

Since sorting performs O(n log n) such comparisons, the actual cost of sorting strings is O(n log n · m), not just O(n log n).
After sorting, comparing adjacent strings takes O(n · m) in the worst case.

Thus, the overall time complexity is:

O(n log n · m)

cjyuan:
Spot on.

illicitonion:

PR #119, opened 2026-02-21
WestMidlands| SDC NOV 2025 |Sara Tahir| Sprint 2 | Improve with caches

SaraTahir28:
## Learners, PR Template

Self checklist

## Changelist

Improved the performance of both the fibonacci and making_change functions by adding caching (memoisation) to avoid recalculating the same subproblems.

## Questions

I dont have any questions.

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,5 +1,5 @@
 from typing import List
-
+cache = {}
 
 def ways_to_make_change(total: int) -> int:
     """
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py

From line 42, we know `coins` can only be one of the following 9 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[1]
[]
```

  - avoid repeatedly creating the same sub-arrays at line 42 (e.g. use another cache), and

SaraTahir28: 1. Thanks for the suggestion . I’ve updated the implementation so I no longer slice the coins list. Instead, I pass a start_index through the recursion, which avoids creating new subarrays entirely.
2. I also switched the cache key to (total, start_index) rather than (total, tuple(coins)). This gives each of the 9 possible sub‑arrays a unique integer “identity” automatically, without needing a separate mapping
3. This keeps the logic the same but removes the repeated list allocations and makes the memoisation faster and cheaper.

cjyuan:
Changes look good.

illicitonion:

PR #120, opened 2026-02-23
WestMidlands| SDC NOV 2025 | Sara Tahir | Sprint 2 |Improve with precomputing

SaraTahir28:
## Learners, PR Template

Self checklist

## Changelist

**Common prefixes**
 In the original implementation we compared every pair of strings, which is O(n²). By sorting the list first, strings with similar prefixes become adjacent, so we only need to compare neighbouring pairs. This reduces the complexity to O(n log n)

**Count Letters**
I have Precomputed lowercase and uppercase character sets to avoid repeated full‑string scans.
This reduces count_letters from O(n²) to O(n).

## Questions

I have no Questions. Thankyou.

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,18 +7,22 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if len(strings)< 2:
+        return ""
+#sorting puts string with similar prefixes next to each other and reduces the complexity from O(n^2)to O(nlogn)
+    strings.sort() 
     longest = ""
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
-            common = find_common_prefix(string, other_string)
-            if len(common) > len(longest):
-                longest = common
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
cjyuan, line 1: The indentation and spacing of the code in this file is a bit off. Can you improve the formatting of the code?

cjyuan (commented):
Can you use complexity to explain why your implementations are better than the originals?

SaraTahir28:
> Can you use complexity to explain why your implementations are better than the originals?

In the original implementation we compared every pair of strings, which is O(n²). By sorting the list first, strings with similar prefixes become adjacent, so we only need to compare neighbouring pairs. This reduces the complexity to O(n log n)

Count Letters
I have Precomputed lowercase and uppercase character sets to avoid repeated full‑string scans.
This reduces count_letters from O(n²) to O(n).

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
SaraTahir28: Thankyou for you suggestion, I  have addressed this in a seperate commit.

cjyuan:
Changes and analysis are good! Well done.

illicitonion:

PR #125, opened 2026-02-27
Sheffield | 25-SDC-Nov | Hassan Osman | Sprint 2 | Implement a linked list in Python

HassanOHOsman:
Self checklist

In this pull request, I implemented a linked list with push_head, pop_tail, and remove operations, all in O(1) time.

### DIFF
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -50,6 +50,9 @@ class LinkedList:
         else:
             self.tail = node.previous
 
+        node.previous = None
+        node.next = None
+
   
 
 
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -29,11 +29,15 @@ class LinkedList:
 
         tail_node = self.tail
         previous = self.tail.previous
+
         self.tail = previous
         if self.tail is not None:
             self.tail.next = None
         else:
             self.head = None
+
+        tail_node.previous = None
+        tail_node.next = None
         
         return tail_node.data
     
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -40,15 +40,18 @@ class LinkedList:
 
 
     def remove(self, node):
+        previous_node = node.previous
+        next_node = node.next
+
         if node.previous is not None:
-            node.previous.next = node.next
+            previous_node.next = next_node
         else:
-            self.head = node.next
+            self.head = next_node
 
         if node.next is not None:
-            node.next.previous = node.previous
+            next_node.previous = previous_node
         else:
-            self.tail = node.previous
+            self.tail = previous_node
 
         node.previous = None
         node.next = None
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -27,17 +27,15 @@ class LinkedList:
         if self.tail is None:
             raise IndexError("Unable to remove from empty linked list")
 
-        #     tail_node = self.tail
-        #     previous = self.tail.previous
-        #     self.tail = previous
-        #     if self.tail is not None:
-        #         self.tail.next = None
-        #     else:
-        #         self.head = None
-        # else:
-            
+        tail_node = self.tail
+        previous = self.tail.previous
+        self.tail = previous
+        if self.tail is not None:
+            self.tail.next = None
+        else:
+            self.head = None
         
-        # return tail_node.data
+        return tail_node.data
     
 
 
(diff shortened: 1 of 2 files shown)


Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,64 @@
...
+            tail_node = self.tail
+            previous = self.tail.previous
+            self.tail = previous
+            if self.tail is not None:
+                self.tail.next = None
+            else:
+                self.head = None
+        else:
+            raise IndexError("Unable to remove from empty linked list")
+
```
cjyuan: - Could consider delegating the node removing task to `remove()` -- less code to maintain.

- Could also consider this style of code writing:

```python
        # Handle all errors and special cases first
        if self.tail is None:
            raise IndexError("Unable to remove from empty linked list")

        # From this point onward, we don't have to worry about special cases.

        # No else needed; one less indentation level.
        tail_node = self.tail
        previous = self.tail.previous
        self.tail = previous
        if self.tail is not None:
            self.tail.next = None
        else:
            self.head = None
```

cjyuan (commented):
The code in `linked_list_test.py` expects both `.next` and `.previous` properties of the removed node to be assigned `None`. Currently your implementation could not pass the tests.

Note: Do you know the why it is a good practice to assign `.next` and `.previous` of the removed node to `None`?

HassanOHOsman:
> The code in `linked_list_test.py` expects both `.next` and `.previous` properties of the removed node to be assigned `None`. Currently your implementation could not pass the tests.
> 
> Note: Do you know the why it is a good practice to assign `.next` and `.previous` of the removed node to `None`?

Strangely, all three tests passed on my end. Here's a screenshot for you reference.
Screenshot 2026-02-27 at 14 41 20

Having said that, I will check it again and see what's gone wrong. Thank you CJ :)

cjyuan:
The original test is not quite complete. I added a few more myself, and here is one of them.
```python
    def test_remove_middle(self):
        l = LinkedList()
        l.push_head("a")
        b = l.push_head("b")
        l.push_head("c")

        l.remove(b)

        self.assertIsNone(b.next)
        self.assertIsNone(b.previous)

        self.assertEqual(l.pop_tail(), "a")
        self.assertEqual(l.pop_tail(), "c")
```

HassanOHOsman:
> The original test is not quite complete. I added a few more myself, and here is one of them.
> 
> ```python
>     def test_remove_middle(self):
>         l = LinkedList()
>         l.push_head("a")
>         b = l.push_head("b")
>         l.push_head("c")
> 
>         l.remove(b)
> 
>         self.assertIsNone(b.next)
>         self.assertIsNone(b.previous)
> 
>         self.assertEqual(l.pop_tail(), "a")
>         self.assertEqual(l.pop_tail(), "c")
> ```

Thanks for sharing, I will update my test file to include what've just shared and  then work to pass it.

HassanOHOsman:
> Note: Do you know the why it is a good practice to assign `.next` and `.previous` of the removed node to `None`?

Is it to ensure that the removed node is dormant/permanently isolated, so that nobody accidentally in the future re-link it back to the list?

cjyuan:
> Is it to ensure that the removed node is dormant/permanently isolated, so that nobody accidentally in the future re-link it back to the list?

That one of the reasons. Another one is related to garbage collection.

cjyuan:
Currently it is still possible that some removed nodes do not have their `.previous` properly set to `None`.

Can you fix the bug?

HassanOHOsman:
updated my logic now so we first capture the previous and next nodes to our current node before any changes we make to the linked array structure (re-linking).

cjyuan:
Currently it is still possible that some removed nodes do not have their `.previous` properly set to `None`.

This is part of the cons of "having more code to maintain" -- focus on where node can be removed from the linked list in your implementation.

HassanOHOsman:
> Currently it is still possible that some removed nodes do not have their `.previous` properly set to `None`.
> 
> This is part of the cons of "having more code to maintain" -- focus on where node can be removed from the linked list in your implementation.

My bad, missed doing the same process inside pop_tail function. Anyway, it's all sorted now. Thanks CJ :)

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,74 @@
...
+        self.tail = previous
+        if self.tail is not None:
+            self.tail.next = None
+        else:
+            self.head = None
+
+        tail_node.previous = None
+        tail_node.next = None
+        
+        return tail_node.data
```
cjyuan, line 32: - Is line 40 needed?

- You could replace the code on lines 30-42 by 3 lines of code if you delegate the node-removing task to `remove()`.

HassanOHOsman: yeah line 40 is definitely useless since a tail node's next should already be None. Not so intelligent of me!

HassanOHOsman: Regarding "You could replace the code on lines 30-42 by 3 lines of code if you delegate the node-removing task to remove()", I just updated my logic inside pop_tail function accordingly.

cjyuan:
Changes look good!

illicitonion:

PR #127, opened 2026-02-27
Manchester | 25-SDC-Nov | Rahwa Haile | Sprint 2 | improve_with_caches

RahwaZeslusHaile:
Manchester | 25-SDC-Nov | Rahwa Haile | Sprint 2 | improve_with_caches
## Learners, PR Template

Self checklist

Description
Optimizes two recursive algorithms using memoization (caching) to dramatically improve performance without changing functionality.

##Changes:

#Fibonacci:

Added cache dictionary to store computed values
Lookup cache before computing, store result after
Reduces time complexity from O(2^n) to O(n)

#Making Change:

Refactored to use coin_index(int) instead of coin list slicing
Added cache with (total, coin_index) key pairs
Avoids redundant subproblem calculations

Please review my work — thank you!

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,32 +1,33 @@
 from typing import List
 
 
+cache = {}
+coins = [200, 100, 50, 20, 10, 5, 2, 1]
+
+
 def ways_to_make_change(total: int) -> int:
-    """
-    Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
 
-    For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
-    """
-    return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
+    cache.clear()
+    return ways_to_make_change_helper(total, 0)
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
-    """
-    Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
-    """
-    if total == 0 or len(coins) == 0:
+def ways_to_make_change_helper(total: int, coin_index: int) -> int:
+    key = (total, coin_index)
+    if key in cache:
+        return cache[key]
+    
+    if total == 0:
+        return 1
+    if coin_index >= len(coins) or total < 0:
         return 0
 
     ways = 0
-    for coin_index in range(len(coins)):
-        coin = coins[coin_index]
-        count_of_coin = 1
-        while coin * count_of_coin <= total:
-            total_from_coins = coin * count_of_coin
-            if total_from_coins == total:
-                ways += 1
-            else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
-                ways += intermediate
-            count_of_coin += 1
+    coin = coins[coin_index]
+    count_of_coin = 0
+    
+    while coin * count_of_coin <= total:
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,33 @@
 from typing import List
 
 
+cache = {}
+coins = [200, 100, 50, 20, 10, 5, 2, 1]
```
cjyuan: If `coins` is to serve as a ***named constant***, consider naming it with all uppercase letters (to follow the naming convention).

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,33 @@
 from typing import List
 
 
+cache = {}
```
cjyuan: Note: Declaring `cache` as a global variable would make the code not thread safe (but we don't have to worry about multi-threaded scenario in this exercise).

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,33 @@
...
-    if total == 0 or len(coins) == 0:
+def ways_to_make_change_helper(total: int, coin_index: int) -> int:
+    key = (total, coin_index)
+    if key in cache:
+        return cache[key]
+    
+    if total == 0:
+        return 1
+    if coin_index >= len(coins) or total < 0:
         return 0
```
cjyuan: Tuple construction and and map/dict lookup operations, while theoretically O(1) operations, are still relatively "slow" compare to simple arithmetic operations.

If you move lines 19-22 before line 15, you may notice slight improvement in performance.

Optional challenge:
And if you can figure out a way to avoid unnecessary lookup **when there is only one coin left to consider**, you shall notice a significant improvement in performance.

cjyuan (commented):
Good job in avoiding repeatedly creating sub-arrays! Almost everyone else missed that.

Code looks good. I only have some suggestions and an optional challenge.

Sprint-2/improve_with_caches/making_change/making_change.py
RahwaZeslusHaile: Got it! @cjyuan I've renamed coins to COINS to follow the constant naming convention. Thanks for catching that!

Sprint-2/improve_with_caches/making_change/making_change.py
RahwaZeslusHaile: Thanks @cjyuan for pointing that out! I understand the thread-safety concern with global variables. Since this is a single-threaded exercise, it's acceptable here.

Sprint-2/improve_with_caches/making_change/making_change.py
RahwaZeslusHaile: @cjyuan Great suggestions! I've implemented both optimizations

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,37 @@
...
-    if total == 0 or len(coins) == 0:
+def ways_to_make_change_helper(total: int, coin_index: int) -> int:
+    if total == 0:
+        return 1
+    if coin_index >= len(COINS) or total < 0:
         return 0
+    
+    # Optimization: if this is the last coin, check if total is divisible by it
+    if coin_index == len(COINS) - 1:
+        return 1 if total % COINS[coin_index] == 0 else 0
```
cjyuan, line 22: Well done.

The check on line 17 is optional because other parts of this function have prevented such condition from happening.

If the objective is to detect empty coin list or negative `total` being passed to `ways_to_make_change()`, we could move the check to that function so that we only need to check the condition once initially.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #130, opened 2026-02-28
Sheffield | 25-SDC-Nov | Hassan Osman | Sprint 2 | Improve code with caches

HassanOHOsman:
Self checklist

In this pull request, I introduced caching (memoisation) to Fibonacci and Making Change directories to make the implementation more efficient.

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -15,7 +15,8 @@ def ways_to_make_change(total: int) -> int:
 
 def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
     key = (total, tuple(coins))
-    
+    cache[key] = ways
+
     if total == 0 or len(coins) == 0:
         return 0
 
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -3,6 +3,7 @@ from typing import List
 cache = {}
 
 
+
 def ways_to_make_change(total: int) -> int:
     """
     Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -29,4 +26,7 @@ def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
                 intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
                 ways += intermediate
             count_of_coin += 1
```
cjyuan: Array creation is a relatively costly operation.

From line 26, we know `coins` can only be one of the following 9 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[]
```

  - avoid repeatedly creating the same sub-arrays at line 26 (e.g. use another cache), and

HassanOHOsman: Done.

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,38 @@
 from typing import List
 
+coins = [200, 100, 50, 20, 10, 5, 2, 1]
```
cjyuan, line 3: - If this is to be used as a constant, best practice is to name it with all uppercase letters.

- We could also keep the `coins` array in the parameter. Arrays are pass by reference to a function so the overhead is negligible.

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,38 @@
...
-    if total == 0 or len(coins) == 0:
+    if key in cache:
+        return cache[key]
+    
+    if total == 0:
+        return 1
+    
+
+    if coin_index == len(coins):
         return 0
```
cjyuan, line 23: Note: Even though key construction and dictionary lookup are O(1) operations, they are still relatively costly than some simple numerical computation or comparisons (code on lines 18 and 22).

This is unrelated to cache, but if you swap the code on lines 18-23 with the code on lines 13-16, you can probably notice some slight improvement in performance.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #132, opened 2026-02-28
London | November 25 | Donara Blanc | Sprint 2 | Improve with caches

donarbl:
Self checklist

## Changelist

both functions now avoid redundant recursive calls

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,22 +1,15 @@
 from typing import List
 
-
 def ways_to_make_change(total: int) -> int:
-    """
-    Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
+    cache = {}
+    return ways_to_make_change_helper(total, (200, 100, 50, 20, 10, 5, 2, 1), cache)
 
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -26,7 +19,12 @@ def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
             if total_from_coins == total:
                 ways += 1
             else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
+                intermediate = ways_to_make_change_helper(
+                    total - total_from_coins,
+                    coins=coins[coin_index + 1:],
+                    cache=cache
+                )
```
cjyuan, line 26: Tuple or array creation are relatively costly operations.

From lines 5 and 24, we know `coins` can only be one of the following 9 tuples:
```
(200, 100, 50, 20, 10, 5, 2, 1)
(100, 50, 20, 10, 5, 2, 1)
(50, 20, 10, 5, 2, 1)
...
(1)
()
```

  - avoid repeatedly creating the same sub-tuples at line 24 (e.g. use another cache), and
  - create key as (total, a_unique_integer_identifying_the_subtuple) instead of as (total, tuple of coins)
     - There are only a small number of different sub-tuples. We can easily assign each sub-tuple a unique integer.

PR #136, opened 2026-03-01
London | 25-SDC-NOV | Jesus del Moral | Sprint 2 | Implement Linked List

delmorallopez:
Implement a linked list in Python

### DIFF
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -1,0 +1,58 @@
+class Node: #A node in a doubly linked list
+    def __init__(self, value): #Initialize a new node with the given value and no next or previous nodes
+        self.value = value
+        self.previous = None
+        self.next = None
+
+
+class LinkedList: # A doubly linked list data structure
+    def __init__(self): # Initialize an empty linked list with no head node
+        self.head = None
+        self.tail = None
+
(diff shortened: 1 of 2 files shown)


Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,58 @@
...
+        value = old_tail.value
+
+        if old_tail.previous:
+          self.tail = old_tail.previous
+          self.tail.next = None
+        else:
+           # Only one element
+           self.head = None
+           self.tail = None
+
```
cjyuan, line 50: - Why not also clean up references in this method? 
 
- Could consider delegating the node removing task to `remove()` -- less code to maintain.

cjyuan (commented):
The PR description is not quite accurate.

Sprint-2/implement_linked_list/linked_list.py
delmorallopez: Delegating to remove() would definitely reduce duplication and keep the logic in one place. It would make the implementation cleaner and easier to maintain, since any changes to the removal logic would only need to be made in a single method.
Delegating to remove(old_tail) does include the cleanup references because remove() method already does this

cjyuan:
All good! Well done.

illicitonion:

PR #137, opened 2026-03-01
London | 25-SDC-NOV | Jesus del Moral | Sprint 2 | LRU Cache

delmorallopez:
LRU Cache

### DIFF
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -1,0 +1,82 @@
+class Node:
+    def __init__(self, key, value):
+        self.key = key
+        self.value = value
+        self.previous = None
+        self.next = None
+
+
+class LruCache:
+    def __init__(self, limit):
+        if limit <= 0:
+            raise ValueError("Limit must be positive")
+
+        self.limit = limit
+        self.cache = {}  # key -> node
+        self.head = None  # Most recently used
+        self.tail = None  # Least recently used
+
+    # ---------------------
+    # Internal helpers
+    # ---------------------
+
+    def _remove_node(self, node):
+        if node.previous:
+            node.previous.next = node.next
+        else:
+            self.head = node.next
+
+        if node.next:
+            node.next.previous = node.previous
+        else:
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
In fact, you can import your LinkedList class from the other exercise, and then store each key-value pairs as a tuple.

Alternatively, you can use `OrderedDict` directly within `LruCache` to maintain order.

delmorallopez:
I updated the implementation using OrderedDict as suggested. This approach simplifies the code and better follows the Single-Responsibility Principle, since order maintenance is handled by OrderedDict rather than a custom linked list.

The cache now: 
- storing key-value pairs in OrderedDict, which preserves insertion order
- moving accessed or updated items to the end to mark them as most recently used
- evicting the least recently used item (the first entry) when the limit is reached
- providing O(1) get and set operations
This removes the need for manual node management and linked list logic, making the code cleaner and easier to maintain while still satisfying LRU behavior.

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,68 @@
+from collections import OrderedDict
+
+class LruCache:
+    def __init__(self, limit):
+        if limit <= 0:
+            raise ValueError("Limit must be positive")
+
+        self.limit = limit
+<<<<<<< HEAD
+        self.cache = {}  # key -> node
```
cjyuan: It seems a merge conflict has not yet been properly resolved.

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,26 @@
...
+
+        self.limit = limit
+        self.cache = OrderedDict()
+
+    def get(self, key):
+        if key not in self.cache:
+            return None
+
+        value = self.cache.pop(key)
+        self.cache[key] = value
```
cjyuan, line 16: OrderedDict has a built-in method that can replace these two operations.

cjyuan (commented):
Changes look good.

illicitonion:

PR #138, opened 2026-03-01
London | 25-SDC-NOV | Jesus del Moral | Sprint 2 | Improve with Caches

delmorallopez:
Fibonacci

Making Change

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,45 +1,52 @@
-from typing import List
+from typing import Dict, Tuple
+
+COINS = [200, 100, 50, 20, 10, 5, 2, 1]
+
 
 def ways_to_make_change(total: int) -> int:
     """
-    Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200,
-    returns a count of all of the ways to make the passed total value.
+    Returns the number of ways to make `total` using UK coin values.
     """
-    cache = {}
-    return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1], cache)
+    cache: Dict[Tuple[int, int], int] = {}
+    return _helper(total, 0, cache)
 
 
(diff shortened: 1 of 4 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -26,7 +33,13 @@ def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
             if total_from_coins == total:
                 ways += 1
             else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
+                intermediate = ways_to_make_change_helper(
+                    total - total_from_coins,
+                    coins=coins[coin_index + 1:],
+                    cache=cache
+                )
```
cjyuan: Array and tuple creations are relatively costly operations.

From lines 9 and 38, we know `coins` can only be one of the following 9 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[1]
[]
```

  - avoid repeatedly creating the same sub-arrays at line 38 (e.g. use another cache), and

delmorallopez: Very good point @cjyuan 

You’re absolutely right that repeatedly slicing the coins list and converting it to a tuple for the cache key introduces unnecessary overhead. Even though there are only 9 possible subarrays, my implementation was recreating them many times during recursion, which adds avoidable allocation cost.

I’ve refactored the solution to:

- Keep a single global COINS list

- Pass a coin_index instead of slicing the list

- Use (total, coin_index) as the cache key instead of (total, tuple(coins))

Since the coin subarrays are deterministic (they’re just suffixes of the original list), the index uniquely identifies each one. This removes repeated list creation and tuple construction, and reduces the constant factor of the algorithm while keeping the same overall time complexity.

cjyuan:
Changes look good. 

Challenge: The code could be further optimized but it is not related to cache. There is a much quicker way to determine number of ways when there is only one coin to consider.

illicitonion:

PR #139, opened 2026-03-01
London | 25-SDC-NOV | Jesus del Moral | Sprint 2 | Improve with Precomputing

delmorallopez:
Common prefix

Count letters

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -20,7 +20,6 @@ def find_longest_common_prefix(strings: List[str]) -> str:
     return longest
 
 
-@lru_cache(maxsize=None)
 def cached_find_common_prefix(left: str, right: str) -> str:
     min_length = min(len(left), len(right))
     for i in range(min_length):
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -1,24 +1,33 @@
 from typing import List
+from functools import lru_cache
 
 
-def find_longest_common_prefix(strings: List[str]):
-    """
-    find_longest_common_prefix returns the longest string common at the start of any two strings in the passed list.
+def find_longest_common_prefix(strings: List[str]) -> str:
+    if len(strings) < 2:
+        return ""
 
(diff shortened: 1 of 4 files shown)


Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
```diff
@@ -1,24 +1,33 @@
...
+        common = cached_find_common_prefix(sorted_strings[i], sorted_strings[i + 1])
+        if len(common) > len(longest):
+            longest = common
+
     return longest
 
 
-def find_common_prefix(left: str, right: str) -> str:
+@lru_cache(maxsize=None)
+def cached_find_common_prefix(left: str, right: str) -> str:
```
cjyuan: Can you justify the use of `@lru_cache`? Do you expect it to always improve the performance?

Sprint-2/improve_with_precomputing/count_letters/count_letters.py
cjyuan, line 1: Very clever approach.

cjyuan (commented):
Can you use complexity to explain how the new implementation is better than the original implementation?

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
delmorallopez: In the original version, we compare every string with every other string. If there are a lot of strings, the number of comparisons grows very quickly because it’s roughly proportional to the square of the number of strings. So doubling the number of strings makes the work grow by about four times.

In the new version, we first sort the strings. Sorting does add some work, but after sorting we only compare neighbouring strings instead of every possible pair. That reduces the number of prefix comparisons dramatically, from “compare everything with everything” to just “compare each string with the next one”.

As a result, the work now grows much more slowly as the number of strings increases. For large inputs, this makes a very noticeable difference in performance compared to the original approach.

So overall, the new implementation scales much better as the list gets bigger.

In this version, @lru_cache stores the result of cached_find_common_prefix(left, right) so that if the same pair of strings is compared again, we can return the result immediately instead of recomputing it.
In this specific implementation, we only compare each adjacent pair once after sorting. That means the same pair of strings is not normally recomputed. So in practice, the cache is unlikely to provide much benefit here.
So I have removed it

cjyuan: Whether lru_cache can help or not really depend on the characteristics of the strings we are comparing. Removing it is optional; it's more important to understand its pros and cons, which you did. Good job!

illicitonion:

PR #141, opened 2026-03-01
London | 25-SDC-Nov | Aida Eslamimoghadam | Sprint 1 | Analyse and refactor functions

aydaeslami:
Self checklist

- [x]My changes meet the requirements of the task

## Changelist

Refactored functions to reduce time complexity by removing nested loops and using sets for faster lookups, improving overall performance.

### DIFF
--- a/Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
+++ b/Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
@@ -10,23 +10,29 @@
  * }
  *
  * Time Complexity:
+ *      ANSWER: Because the function loops through the numbers array once,and the running time increases linearly with the size of the input.
+ *              O(n). 
  * Space Complexity:
+ *      ANSWER: O(1), Because the function only uses two extra variables (sum and product) and does not create any additional data structures based on the input size.
+ * 
+ 
  * Optimal Time Complexity:
+            ANSWER: This is optimal because every number in the array must be processed at least once to calculate the sum and product, so it cannot be done faster than O(n).
+            O(n) .
+
  *
  * @param {Array<number>} numbers - Numbers to process
  * @returns {Object} Object containing running total and product
  */
 export function calculateSumAndProduct(numbers) {
   let sum = 0;
+  let product = 1;
+  
+ 
   for (const num of numbers) {
     sum += num;
-  }
-
-  let product = 1;
-  for (const num of numbers) {
     product *= num;
   }
-
   return {
     sum: sum,
     product: product,
--- a/Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
+++ b/Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
@@ -20,6 +20,8 @@
             ANSWER: This is optimal because every number in the array must be processed at least once to calculate the sum and product, so it cannot be done faster than O(n).
             O(n) .
 
+
+   Original Time Complexity: O(2n)
  *
  * @param {Array<number>} numbers - Numbers to process
  * @returns {Object} Object containing running total and product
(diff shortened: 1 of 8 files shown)


Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -2,13 +2,23 @@
  * Finds common items between two arrays.
  *
  * Time Complexity:
- * Space Complexity:
+ *        ANSWER: O(n + m)
+          Because the second array is converted into a Set once, and the first array is looped through once.
+
```
cjyuan, line 7: I think "Time Complexity" here refers to the time complexity of the original implementation, and "Optimal Time Complexity" refers to the time complexity of the refactored version. (Otherwise they will always have the same complexity)

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -2,13 +2,23 @@
...
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
-export const findCommonItems = (firstArray, secondArray) => [
-  ...new Set(firstArray.filter((item) => secondArray.includes(item))),
-];
+export const findCommonItems = (firstArray, secondArray) => {
+  const secondSet = new Set(secondArray);
+
+  return [...new Set(firstArray.filter(item => secondSet.has(item)))];
```
cjyuan: Note: Set object has a built-in method that can be used to find the "common items" between two sets.
Using built-in method usually means better performance and shorter code.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -14,8 +14,10 @@ def find_common_items(
...
     common_items: List[ItemType] = []
+    second_sequence_set = set(second_sequence)
+
     for i in first_sequence:
-        for j in second_sequence:
-            if i == j and i not in common_items:
-                common_items.append(i)
-    return common_items
+        if i in second_sequence_set and i not in common_items:
+            common_items.append(i)
```
cjyuan: Can also consider using a built-in set operation for better performance and simpler code.

cjyuan (commented):
Code is good.

Can you also show the time complexity of the original implementation?

Sprint-1/Python/find_common_items/find_common_items.py
aydaeslami: Thanks for your feedback @cjyuan !I’ve made the suggested changes, including using the built-in set intersection and updating the complexity comments. Please let me know if anything else needs improvement.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -9,13 +9,8 @@ def find_common_items(
     """
     Find common items between two arrays.
 
-    Time Complexity:
-    Space Complexity:
-    Optimal time complexity:
+    Time Complexity: O(n + m) process both sequences
```
cjyuan: The complexity of the original implementation is not O(n+m).

cjyuan (commented):
Can you also show the time complexity of the original implementation (in each file)?

Sprint-1/Python/find_common_items/find_common_items.py
cjyuan: The new implementation looks good.

Sprint-1/Python/find_common_items/find_common_items.py
aydaeslami: Yes, You are right, Thank you for pointing that out. I've corrected the time complexity

Sprint-1/JavaScript/calculateSumAndProduct/calculateSumAndProduct.js
```diff
@@ -20,6 +20,8 @@
             ANSWER: This is optimal because every number in the array must be processed at least once to calculate the sum and product, so it cannot be done faster than O(n).
             O(n) .
 
+
+   Original Time Complexity: O(2n)
```
cjyuan, line 24: In complexity analysis, O(2n) is equivalent to O(n). I think O(n) is the standard notation.

cjyuan (commented):
Looks good.

illicitonion:

PR #142, opened 2026-03-01
London | SDC Nov 25 | Emiliano Uruena | Sprint 2 |Improve with precomputing

Emilianouz:
## Learners, PR Template

Self checklist

## Changelist

Refactor common_prefix and count_letter functions with precomputing.

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,12 +7,17 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if len(strings) < 2:
+        return ""
+    # reducing comparison by sorting places strings with similar prefixes next to each other...
+    strings.sort()
+
     longest = ""
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
-            common = find_common_prefix(string, other_string)
-            if len(common) > len(longest):
-                longest = common
+    for i in range(len(strings) - 1):
+        # compares only adjacent strings after sorting
+        common = find_common_prefix(strings[i], strings[i + 1])
+        if len(common) > len(longest):
+            longest = common
     return longest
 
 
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
```diff
@@ -7,12 +7,17 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if len(strings) < 2:
+        return ""
+    # reducing comparison by sorting places strings with similar prefixes next to each other...
+    strings.sort()
```
cjyuan: `strings.sort()` can modify the original list passed into the function by the caller.
How can you prevent this side effect?

cjyuan (commented):
Can you use complexity to explain how the new implementation are better than the original implementation?

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
Emilianouz: Good point... strings.sort() modifies the original list in place. I’ll use sorted(strings) to prevent the side effect.

cjyuan:
Can you also address this comment: https://github.com/CodeYourFuture/Module-Complexity/pull/142#pullrequestreview-3873696181

Emilianouz:
> Can you use complexity to explain how the new implementation are better than the original implementation?

@cjyuan  The original had complexity O(n² × m) because compares every possible pair of strings, and each comparison can take up to O(string length). The improved version first sorts the list in O(n log n) then only compares neighbours strings  O(n × m). That reduces the number of comparisons.

cjyuan:
If we are factoring in the length of the strings, m, then the complexity of sorting won't just be O(nlogn). nlogn measures only the number of comparisons needed.

Emilianouz:
> If we are factoring in the length of the strings, m, then the complexity of sorting won't just be O(nlogn). nlogn measures only the number of comparisons needed.

The new implementation first sorts the strings, which costs O(n log n × m) since string comparisons can take up to O(m). After sorting, only adjacent strings are compared, reducing the comparison phase to O(n × m).

Overall the complexity improves from O(n² × m) to O(n log n × m)

cjyuan:
All good. Well done.

illicitonion:

PR #143, opened 2026-03-01
London | 25-SDC-Nov | Aida Eslamimoghadam | Sprint 2 | Implement LRU cache

aydaeslami:
Self checklist

## Changelist
Implemented LRU cache with O(1) get and set operations using a doubly linked list and dictionary.

### DIFF
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -1,0 +1,70 @@
+class Node:
+    def __init__(self, key, value):
+        self.key = key
+        self.value = value
+        self.previous = None
+        self.next = None
+
+
+class LruCache:
+    def __init__(self, limit):
+        if limit <= 0:
+            raise ValueError("Limit must be greater than 0")
+
+        self.limit = limit
+        self.cache = {}
+        self.head = None
+        self.tail = None
+
+    def _remove(self, node):
+        if node.previous:
+            node.previous.next = node.next
+        else:
+            self.head = node.next
+
+        if node.next:
+            node.next.previous = node.previous
+        else:
+            self.tail = node.previous
+
+        node.previous = None
+        node.next = None
+
+    def _add_to_head(self, node):
+        node.next = self.head
+        node.previous = None
+
+        if self.head:
+            self.head.previous = node
+        else:
+            self.tail = node
+
+        self.head = node
+
+    def get(self, key):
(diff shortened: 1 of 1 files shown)


cjyuan (commented):
In fact, you can import your LinkedList class from the other exercise, and then store each key-value pairs as a tuple.

Alternatively, you can use `OrderedDict` directly within `LruCache` to maintain order.

aydaeslami:
Thank you @cjyuan for the suggestion. I refactored the implementation to use OrderedDict , and it made the code much cleaner and easier to read.

I also appreciated the reminder about the Single Responsibility Principle (SRP).

Thanks for the helpful feedback!

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,46 @@
...
+
+        self.limit = limit
+        self.cache = OrderedDict()
+
+    def get(self, key):
+        if key not in self.cache:
+            return None
+
+        value = self.cache.pop(key)
+        self.cache[key] = value
```
cjyuan: Does OrderedDict have any built-in method to move its item to the end?

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,46 @@
...
+cache.set("Clare", 3)
+
+print(cache.get("Aida")) 
+
+cache.set("David", 4)
+
+print(cache.get("Bob")) 
+print(cache.get("Clare")) 
+print(cache.get("David"))  
+
```
cjyuan: Why not implement this in `lru_cache_test.py`?

Sprint-2/implement_lru_cache/lru_cache.py
aydaeslami: Thanks for the suggestion. I've replaced the pop-and-reinsert approach with `OrderedDict.move_to_end()`, which makes the code cleaner and more readable.

I've also run the full test suite again, and all tests are passing.

Please let me know if there's anything else I should change or improve.

cjyuan:
All good.

Sprint-2/implement_lru_cache/lru_cache.py
cjyuan: It could also be used in the set method.

illicitonion:

PR #144, opened 2026-03-01
London | 25-SDC-Nov | Aida Eslamimoghadam | Sprint 2 | Add caching to improve performance

aydaeslami:
Self checklist

## Changelist

Added caching (memoisation) to avoid repeated calculations and improve performance of recursive functions.

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,20 +1,21 @@
 from typing import List
 
+cache = {}
 
 def ways_to_make_change(total: int) -> int:
-    """
-    Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
-
-    For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
-    """
     return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
 
 
 def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
-    """
-    Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
-    """
-    if total == 0 or len(coins) == 0:
+    key = (total, tuple(coins))
+
+    if key in cache:
+        return cache[key]
+
+    if total == 0:
+        return 1
+
+    if total < 0 or len(coins) == 0:
         return 0
 
     ways = 0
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -26,7 +27,11 @@ def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
             if total_from_coins == total:
                 ways += 1
             else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
-                ways += intermediate
+                ways += ways_to_make_change_helper(
+                    total - total_from_coins,
+                    coins=coins[coin_index + 1:]
+                )
```
cjyuan: Array and tuple creations are relatively costly operations.

From lines 6 and 32, we know `coins` can only be one of the following 9 arrays:
```
[100, 50, 20, 10, 5, 2, 1]
[50, 20, 10, 5, 2, 1]
...
[1]
[]
```

  - avoid repeatedly creating the same sub-arrays at line 32 (e.g. use another cache), and

aydaeslami:
Thanks a million @cjyuan for your suggestion, I updated the solution to use start_index instead of creating coin sub-arrays, and changed the cache key to total, start_index to avoid unnecessary array and tuple creation.
Please let me know if there is anything that I should change.

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,26 +1,27 @@
...
 
 def ways_to_make_change(total: int) -> int:
-    return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
+    return ways_to_make_change_helper(total, 0)
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
-    key = (total, tuple(coins))
+def ways_to_make_change_helper(total: int, start_index: int) -> int:
+    key = (total, start_index)
```
cjyuan, line 11: Why not pass the coin list through the parameter (in addition to passing `start_index`) so that the function can be reused for different coin list?

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,26 +1,27 @@
...
         return 1
 
-    if total < 0 or len(coins) == 0:
+    if total < 0 or start_index >= len(COINS):
         return 0
 
     ways = 0
-    for coin_index in range(len(coins)):
-        coin = coins[coin_index]
+    for coin_index in range(start_index, len(COINS)):
```
cjyuan, line 23: This loop is actually redundant. Subsequent coins will be considered in the recursive function calls anyway.

PR #145, opened 2026-03-01
London | 25-SDC-Nov | Aida Eslamimoghadam | Sprint 2 | Use precomputing for better performance

aydaeslami:
Self checklist

## Changelist

Introduced precomputing to reduce repeated work and improve performance of common_prefix and count_letters functions.

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -7,12 +7,19 @@ def find_longest_common_prefix(strings: List[str]):
 
     In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
     """
+    if len(strings) < 2:
+        return ""
+
+    strings = sorted(strings)
     longest = ""
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Code looks good.

Can you use complexity to explain how the new implementation is better than the original implementation?

aydaeslami:
Thank you, @cjyuan, for your help and support.

For " find_longest_common_prefix" : 
The previous implementation compared the common prefix of every pair of strings, resulting in a time complexity of O(n² × m). The new implementation sorts the strings first and only compares adjacent strings, reducing the time complexity to O(n log n + n × m).

And 

For " count_letters" : 
The previous implementation searched the string for each uppercase letter, resulting in a time complexity of O(n²). The new implementation uses sets for faster lookups, reducing the time complexity to O(n).

cjyuan:
Sorting strings involve comparing strings character by character. If we take into account, $m$, the average length of the strings in the given array, then the time complexity of sorting would also affected by $m$.

aydaeslami:
Thanks @cjyuan, Yes, you are right. i missed the cost of string comparison during sorting. 
ِThe overal complexity is :O(n log n × m + n × m)
simplifies to : O(n log n × m)

This part is a bit complicated for me. Please let me know if I need to change anything.

Many thanks.

cjyuan:
All good. Nothing needs to be changed.

PR #146, opened 2026-03-01
London | 25-SDC-Nov | Aida Eslamimoghadam | Sprint 2 | Implement a linked list in python

aydaeslami:
Self checklist

## Changelist
Implemented a doubly linked list supporting push_head, pop_tail, and remove operations in O(1) time.

### DIFF
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -16,12 +16,10 @@ class LinkedList:
         new_node.next = self.head
         if self.head:
             self.head.previous = new_node
-
-        self.head = new_node
-
-        if self.tail is None:
+        else:
             self.tail = new_node
 
+        self.head = new_node
         return new_node
 
     def pop_tail(self):
@@ -44,6 +42,8 @@ class LinkedList:
             node.previous.next = node.next
         else:
             self.head = node.next
+            if self.head:
+                self.head.previous = None
 
         if node.next:
             node.next.previous = node.previous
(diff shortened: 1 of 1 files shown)


Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,51 @@
...
+            return None
+
+        removed = self.tail
+
+        if self.tail.previous:
+            self.tail = self.tail.previous
+            self.tail.next = None
+        else:
+            self.head = None
+            self.tail = None
```
cjyuan: Could consider delegating the node removing task to `remove()` -- less code to maintain.

cjyuan (commented):
The code in `linked_list_test.py` expects both `.next` and `.previous` properties of the removed node to be assigned `None`. Currently your implementation could not pass the tests.

Note: Do you know the why it is a good practice to assign `.next` and `.previous` of the removed node to `None`?

aydaeslami:
Thanks @cjyuan, for your helpful feedback and suggestions.

My tests were passing even without setting `node.previous` and `node.next` to `None`, so I didn't realize this issue was there. I've now added those lines and updated the implementation accordingly.

**Do you know the why it is a good practice to assign .next and .previous of the removed node to None**

_I think it's a good practice because if we still have access to the removed node, we could accidentally reach other nodes through its .next and .previous links. Setting them to None makes it clear that the node is no longer part of the list and helps avoid mistakes._

Thanks again for taking the time to review it and point it out.

cjyuan:
Changes look good.

I forgot to suggest looking into `__slots__` during my previous review. Could you spend a little time reading up on it and how it can be used in class design? No changes are expected; this is just for learning purposes.

illicitonion:

PR #149, opened 2026-03-05
London | 25-SDC-July | Eyuel Abraham | Sprint 2 | Improve with caches 

eyuell21:
## Learners, PR Template

Self checklist

## Changelist

Added memoization to improve performance.

## Questions

 No Questions.

### DIFF
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,4 +1,8 @@
-def fibonacci(n):
+def fibonacci(n, memo={}):
+    if n in memo:
+        return memo[n]
     if n <= 1:
         return n
-    return fibonacci(n - 1) + fibonacci(n - 2)
+    
+    memo[n] = fibonacci(n - 1, memo) + fibonacci(n - 2, memo)
+    return memo[n]
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,8 +1,10 @@
-def fibonacci(n, memo={}):
-    if n in memo:
-        return memo[n]
+_memo = {}
+
+def fibonacci(n):
+    if n in _memo:
+        return _memo[n]
     if n <= 1:
         return n
     
-    memo[n] = fibonacci(n - 1, memo) + fibonacci(n - 2, memo)
-    return memo[n]
+    _memo[n] = fibonacci(n - 1) + fibonacci(n - 2)
+    return _memo[n]
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,5 +1,4 @@
-from typing import List
-
+from typing import List, Dict, Tuple
 
 def ways_to_make_change(total: int) -> int:
     """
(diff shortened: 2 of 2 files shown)


Sprint-2/improve_with_caches/fibonacci/fibonacci.py
```diff
@@ -1,4 +1,8 @@
-def fibonacci(n):
+def fibonacci(n, memo={}):
```
SlideGauge: Here is a classic python trap: default argument is not recreated, but is used all the type (by Python design).
so it means if I do not pass a memo argument, additions happening to it will be stored between  function calls (due to argument's value referring to the same instance of the memo dict.
Whilst specifically in this case (i.e. in fibonacci), this  behaviour is to some extent good, I'd advice to do this explicitly instead:
1. make memo as a global variable (internal usage, i.e var with a  prefix "_"). This is one of the cases when global var is okeish 
2. remove memo as an argument of the function. From perspective of me, as a user of the function, the memo arg only  confuse me (like "should I pass it? If  yes, what should I pass" and so on)

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,41 @@
...
         while coin * count_of_coin <= total:
-            total_from_coins = coin * count_of_coin
-            if total_from_coins == total:
+            remainder = total - coin * count_of_coin
+            if remainder == 0:
                 ways += 1
             else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
-                ways += intermediate
+                ways += ways_to_make_change_helper(remainder, coins[coin_index + 1:], memo)
```
SlideGauge, line 41: Overall good approach with memoization, but I have one tiny request here:
What happens when we do slicing in Python? (i.e. do varOfTypeList[x:y])?  Does it allocate additional memory, potentially a lot of? Can we avoid additional allocation  of a lot of memory?

SlideGauge (commented):
Overall decent approaches with solving via caching/memoization. I've got a couple of requests, which will polish both solutions, could you fulfil them?

Sprint-2/improve_with_caches/making_change/making_change.py
eyuell21: In Python, doing list[x:y] creates a new list containing the elements from index x to y-1. This requires allocating additional memory proportional to the slice length. we can use a module called intertools (itertools.islice), which returns a lazy iterator over that range instead of creating a new list to avoid excess memory allocation.

eyuell21:
Hello @SlideGauge, I've made the requested changes.

eyuell21:
Hello! @illicitonion Can I get a review on this one please? Thank you!

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,41 @@
...
     For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
     """
-    return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
+    coins = [200, 100, 50, 20, 10, 5, 2, 1]
+    memo: Dict[Tuple[int, int], int] = {}
+    return ways_to_make_change_helper(total, coins, memo)
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+def ways_to_make_change_helper(total: int, coins: List[int], memo: Dict[Tuple[int, int], int]) -> int:
```
illicitonion, line 14: I'd probably document what the cache key here is in the docstring - it's not obvious what the two values in the tuple are (and it took me a minute to work out why `len(coins)` was a valid thing to include in the cache key at all).

illicitonion (commented):
Generally looks good - one documentation question

illicitonion:

PR #155, opened 2026-04-16
London | 26-SDC-Mar | Ebrahim Beiati-Asl | Sprint 2 | implement_skip_list

ebrahimbeiati:
## Self checklist

### DIFF
--- a/Sprint-2/implement_skip_list/skip_list.py
+++ b/Sprint-2/implement_skip_list/skip_list.py
@@ -1,0 +1,72 @@
+import random
+
+
+class Node:
+    def __init__(self, height, value):
+        self.value = value
+        self.forward = [None] * height
+
+
+class SkipList:
+    MAX_HEIGHT = 16
+    P = 0.5
+
+    def __init__(self):
+        self.head = Node(self.MAX_HEIGHT, None)
+        self.level = 1
+        self.length = 0
+
+    def __len__(self):
+        return self.length
+
+    def random_height(self):
+        height = 1
+        while random.random() < self.P and height < self.MAX_HEIGHT:
+            height += 1
+        return height
+
+    def _find_updated_path(self, value):
+        update = [None] * self.MAX_HEIGHT
+        current = self.head
+        for i in range(self.level - 1, -1, -1):
+            while current.forward[i] and current.forward[i].value < value:
+                current = current.forward[i]
+            update[i] = current
+        return update
+
+    def __contains__(self, value):
+        current = self.head
+
+        for i in range(self.level - 1, -1, -1):
+            while current.forward[i] and current.forward[i].value < value:
+                current = current.forward[i]
+
+        current = current.forward[0]
+        return current is not None and current.value == value
+
+    def insert(self, value):
+        update = self._find_updated_path(value)
+        forward_node = update[0].forward[0]
+
+        if forward_node and forward_node.value == value:
+            return
+        node_height = self.random_height()
+        if node_height > self.level:
+            for i in range(self.level, node_height):
+                update[i] = self.head
+            self.level = node_height
+
+        new_node = Node(node_height, value)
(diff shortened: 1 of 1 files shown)


Sprint-2/implement_skip_list/skip_list.py
```diff
@@ -0,0 +1,66 @@
...
+                return j
+        return len(self.data)
+
+    def insert(self, value):
+        """Insert value into the skip list, maintaining sorted order."""
+        pos = self._find_position(value)
+
+        if pos < len(self.data) and self.data[pos] == value:
+            return  # value already exists, do not insert duplicates
+        self.data.insert(pos, value)
```
cjyuan: This insertion operation is O(n), which does not meet the requirement.

cjyuan (commented):
What you implemented is not quite a typical SkipList. In particular, your insertion operation is O(n) whereas the insertion cost of a SkipList is O(log n).

A SkipList should be linked list-liked data structure.

Sprint-2/implement_skip_list/skip_list.py
```diff
@@ -0,0 +1,128 @@
...
+        self.skips = []  # list of skip pointers, each is a tuple (index, value)
+
+    def _get_node(self, index):
+        """Get the node at the given index."""
+        current = self.head
+        for _ in range(index):
+            if current is None:
+                return None
+            current = current.next
+        return current
```
cjyuan: This function performs a linear traversal on the linked list from head.

Sprint-2/implement_skip_list/skip_list.py
```diff
@@ -0,0 +1,128 @@
...
+        if self.length == 0:
+            return 0
+
+        # Use skip pointers to find the range where value should be
+        for i in range(len(self.skips) - 1):
+            a = self.skips[i]
+            b = self.skips[i + 1]
+
+            node_a = self._get_node(a)
+            node_b = self._get_node(b)
```
cjyuan: This loop, together with the linear traversal performed by `_get_node(pos)`, suggests a complexity of worse than O(n).

cjyuan (commented):
What you implemented does not look like a "Skip List".

May I suggest search for Skip List pseudo code, and then convert the pseudo code to Python? There may be more than one way to implement Skip List, so look for something you can understand.

This video gives a good illustration of how Skip List works, but it does not include any pseudo code: https://www.youtube.com/watch?v=UGaOXaXAM5M

Sprint-2/implement_skip_list/skip_list.py
cjyuan: When index is close to `n`, this function has to traverse about `n` nodes to reach current. 

The complexity of a function calling this function will be at least O(n).

cjyuan (commented):
Your implementation does not look like a typical Skip List. I don't think you can make it a Skip List by making some changes. 

I would suggest try to understand how Skip List works first, then study the pseudo code of a Skip List, and then reimplement it from the beginning.

Sprint-2/implement_skip_list/skip_list.py
```diff
@@ -0,0 +1,72 @@
...
+        self.length = 0
+
+    def __len__(self):
+        return self.length
+
+    def random_height(self):
+        height = 1
+        while random.random() < self.P and height < self.MAX_HEIGHT:
+            height += 1
+        return height
```
cjyuan, line 26: Could consider follow naming convention to make this a "private" method because it is only for internal use.

cjyuan (commented):
Looks good.

Which sources did you use as references?

Code does not have any comment. Do you think you'll be able to explain every piece of code you wrote three months from now? If not, it's a good practice to add comments to remind yourself what the code does and why it was written that way.

illicitonion:

PR #161, opened 2026-04-18
Scotland | SDC NOV | Katarzyna Kazimierczuk | Sprint 2 | Improve with caches

katarzynakaz:
# Learners, PR Template

Self checklist

## Changelist

Improve time complexity using memoization for recursive functions.

### DIFF
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,12 +1,12 @@
-def fibonacci(n, cache=None):
-    if cache is None:
-        cache = {}
+def fibonacci(n, memo=None):
+    if memo is None:
+        memo = {}
 
-    if n in cache:
-        return cache[n]
+    if n in memo:
+        return memo[n]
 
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/fibonacci/fibonacci.py
```diff
@@ -1,4 +1,12 @@
-def fibonacci(n):
+def fibonacci(n, memo=None):
```
illicitonion, line 1: This works, and will cache all of the calls within one call to `fibonacci`, but if we wrote this code:

```python
fibonacci(20)
fibonacci(20)
```

The second call wouldn't be any faster than the first. Can you think how to get this benefit as well?

illicitonion (commented):
Looks good, well done!

Sprint-2/improve_with_caches/fibonacci/fibonacci.py
katarzynakaz: Yes, it would have to persist outside of the function to reuse the value.

illicitonion:

PR #164, opened 2026-06-09
NW | 26-SDC-Mar | TzeMing Ho | Sprint 1 | Analyse and Refactor Functions

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

I have attempted the exercises with comments to explain in big O, and tried to optimize the function if possible.

### DIFF
--- a/Sprint-1/Python/find_common_items/find_common_items.py
+++ b/Sprint-1/Python/find_common_items/find_common_items.py
@@ -20,8 +20,9 @@ def find_common_items(
     #             common_items.append(i)
     # return common_items
 
-    firstSet = {*first_sequence}
+    first_set = set(first_sequence)
 
-    commonInSecond = filter(lambda item: item in firstSet, second_sequence)
+    common_in_second = filter(lambda item: item in first_set, second_sequence)
 
-    return list({*commonInSecond})
+    return list(set(common_in_second))
+
--- a/Sprint-1/Python/find_common_items/find_common_items.py
+++ b/Sprint-1/Python/find_common_items/find_common_items.py
@@ -9,13 +9,19 @@ def find_common_items(
     """
     Find common items between two arrays.
 
-    Time Complexity:
-    Space Complexity:
-    Optimal time complexity:
+    Time Complexity: O(N*M), it is a loop inside a loop
+    Space Complexity: O(N), the size of the common_items would grow to N in the worst case
+    Optimal time complexity: O(N + M), N is for the firstSet, M is the filter in second
     """
-    common_items: List[ItemType] = []
-    for i in first_sequence:
-        for j in second_sequence:
-            if i == j and i not in common_items:
-                common_items.append(i)
-    return common_items
+    # common_items: List[ItemType] = []
+    # for i in first_sequence:
+    #     for j in second_sequence:
+    #         if i == j and i not in common_items:
+    #             common_items.append(i)
+    # return common_items
+
+    firstSet = {*first_sequence}
+
+    commonInSecond = filter(lambda item: item in firstSet, second_sequence)
+
+    return list({*commonInSecond})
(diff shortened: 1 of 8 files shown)


Sprint-1/Python/has_pair_with_sum/has_pair_with_sum.py
```diff
@@ -7,12 +7,22 @@ def has_pair_with_sum(numbers: List[Number], target_sum: Number) -> bool:
...
+    # for i in range(len(numbers)):
+    #     for j in range(i + 1, len(numbers)):
+    #         if numbers[i] + numbers[j] == target_sum:
+    #             return True
+    # return False
+
+    targetSet = set()
+
+    for num in numbers:
+        targetNum = target_sum - num
```
cjyuan: Note: ***snake_case*** is the standard convention for variables and functions in Python.

Sprint-1/Python/remove_duplicates/remove_duplicates.py
```diff
@@ -7,19 +7,28 @@ def remove_duplicates(values: Sequence[ItemType]) -> List[ItemType]:
...
-            if value == existing:
-                is_duplicate = True
-                break
-        if not is_duplicate:
+        if value not in seen:
+            seen.add(value)
             unique_items.append(value)
-
+    
     return unique_items
```
cjyuan: Why not use the same approach you used in `Sprint-1/JavaScript/removeDuplicates/removeDuplicates.mjs`?

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -9,13 +9,19 @@ def find_common_items(
...
+    #     for j in second_sequence:
+    #         if i == j and i not in common_items:
+    #             common_items.append(i)
+    # return common_items
+
+    firstSet = {*first_sequence}
+
+    commonInSecond = filter(lambda item: item in firstSet, second_sequence)
+
+    return list({*commonInSecond})
```
cjyuan: Could also use the built-in Set intersect operation.

cjyuan (commented):
Looks good. Well done.

Sprint-1/Python/remove_duplicates/remove_duplicates.py
TzeMingHo: I tried to return list(set(values)). I thought it would have been the same as JavaScript. However, it can't pass the test, as the order of the list is different from the test case. So, I had to add the element to the list one by one.

cjyuan: Right. Python Set does not preserve order. There is, however, another built-in Python class that we can use to produce a single line solution.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -20,8 +20,9 @@ def find_common_items(
     #             common_items.append(i)
     # return common_items
 
-    firstSet = {*first_sequence}
+    first_set = set(first_sequence)
 
-    commonInSecond = filter(lambda item: item in firstSet, second_sequence)
+    common_in_second = filter(lambda item: item in first_set, second_sequence)
+
+    return list(set(common_in_second))
```
cjyuan: I meant, using this https://docs.python.org/3/library/stdtypes.html#frozenset.intersection
And there is even an overridden operator, `&`, for set intersection.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -20,8 +20,9 @@ def find_common_items(
     #             common_items.append(i)
     # return common_items
 
-    firstSet = {*first_sequence}
+    first_set = set(first_sequence)
```
cjyuan: Both of these syntaxes produces set.

illicitonion:

PR #168, opened 2026-06-16
NW | 26-SDC-Mar | TzeMing Ho | Sprint 2 | implement_linked_list

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

1. Created a class Node
2. Created a class LinkedList
3. Added methods: push_head, pop_tail, remove, and push_tail

### DIFF
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -1,0 +1,77 @@
+class Node:
+    def __init__(self, value):
+        self.value = value
+        self.next = None
+        self.previous = None
+
+class LinkedList:
+    def __init__(self):
+        self.head = None
+        self.tail = None
+
+    def push_head(self, value):
+        new_node = Node(value)
+
+        if self.head == None:
+            self.head = new_node
+            self.tail = new_node
+        else:
+            new_node.next = self.head
+            self.head.previous = new_node
+            self.head = new_node
+        
+        return new_node
+    
+    def pop_tail(self):
+
+        if self.tail == None:
+            return
+        elif self.tail == self.head:
+            val = self.tail.value
+            self.tail = None
+            self.head = None
+            return val
+        else:
+            val = self.tail.value
+            new_tail = self.tail.previous
+            new_tail.next = None
+            self.tail.previous = None
+            self.tail = new_tail
+            return val
+
+    def remove(self, Node):
+        if self.head == None and self.tail == None:
+            return
+        
+        new_previous = Node.previous
+        new_next = Node.next
+
+        if self.head == Node and self.tail == Node:
+            Node.value = None
+            self.head = None
+            self.tail = None
+        elif self.head == Node :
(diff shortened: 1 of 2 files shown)


Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,77 @@
...
+            return val
+        else:
+            val = self.tail.value
+            new_tail = self.tail.previous
+            new_tail.next = None
+            self.tail.previous = None
+            self.tail = new_tail
+            return val
+
+    def remove(self, Node):
```
cjyuan: Python naming convention for parameters/variables is lower snake case

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,77 @@
...
+        if self.head == None and self.tail == None:
+            return
+        
+        new_previous = Node.previous
+        new_next = Node.next
+
+        if self.head == Node and self.tail == Node:
+            Node.value = None
```
cjyuan: Why clear the value of the removed node?
If it is needed, why not do the same when the removed node is not the only node in the linked list?

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,77 @@
...
+            self.head = None
+            self.tail = None
+        elif self.head == Node :
+            Node.previous = None
+            Node.next = None
+            new_next.previous = None
+            self.head = new_next
+        elif self.tail == Node:
+            self.pop_tail()
```
cjyuan: Could consider writing the "remove tail" code in this function, and then use `remove()` in `pop_tail()`.
- Have all the remove logic in one place.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,77 @@
...
+        elif new_next and new_previous:
+            new_previous.next = new_next
+            new_next.previous = new_previous
+            Node.previous = None
+            Node.next = None
```
cjyuan: Could probably factor out these two lines of code.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,69 @@
...
+    def pop_tail(self):
+
+        if self.tail == None:
+            return None
+
+        return self.remove(self.tail)
+
+    def remove(self, node):
+        if self.head == None and self.tail == None:
+            return
```
cjyuan: Could use the same syntax to return `None`.

cjyuan (commented):
Changes look good.

Code could probably use some internal documentation.

cjyuan:
Forgot to suggest researching this

> When defining a class, how to use `__slots__` to restrict which attributes its instance can have in order to reduce per-instance memory usage?

illicitonion:

PR #169, opened 2026-06-17
NW | 26-SDC-Mar | TzeMing Ho | Sprint 2 | implement_lru_cache

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

1. Borrowed the linked list from the previous exercise and made some updates 
  1.1 Accepts a key, a value in the class Node
  1.2 Methods in class LinkedList accept Node
3. Created class LruCache
4. Created methods: get, set, and a helper function move_to_front in class LruCache
5. Added slots for classes

### DIFF
--- a/Sprint-2/implement_lru_cache/linked_list.py
+++ b/Sprint-2/implement_lru_cache/linked_list.py
@@ -1,4 +1,6 @@
 class Node:
+    __slots__ = ("key", "value", "previous", "next")
+
     def __init__(self, key, value):
         self.key = key
         self.value = value
@@ -6,6 +8,8 @@ class Node:
         self.next = None
 
 class LinkedList:
+    __slots__ = ("head", "tail")
+
     def __init__(self):
         self.head = None
         self.tail = None
@@ -33,7 +37,7 @@ class LinkedList:
 
     def remove(self, node):
         if self.head == None and self.tail == None:
-            return
+            return None
         
         new_previous = node.previous
         new_next = node.next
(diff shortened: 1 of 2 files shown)


Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,29 @@
...
+            return None
+        else:
+            most_recent_node = self.linked_list.remove(self.dict[key])
+            self.linked_list.push_head(most_recent_node)
+            return most_recent_node.value
+        
+    def set(self, key, value):
+        if key in self.dict:
+            self.dict[key].value = value
+            self.get(key)
```
cjyuan: With the current approach of moving a node to the front:
- It is not immediately obvious why this method calls `get()`
- Calling `get()` introduces an additional (but unnecessary) dictionary lookup

Could you explore a cleaner way to structure the code?

Sprint-2/implement_lru_cache/linked_list.py
```diff
@@ -0,0 +1,73 @@
+class Node:
+    __slots__ = ("key", "value", "previous", "next")
+
+    def __init__(self, key, value):
+        self.key = key
+        self.value = value
```
cjyuan, line 6: Making each node to keep two values (`key` and `value`) would make the linked list less attractive (because most uses of linked list only needs to store one value in each node).

Could you describe how you would implement the LRU cache if you were to import your linked list implemented in the other exercise (unmodified)? 

**No change needed**

cjyuan (commented):
Have you also looked up "Best practices about using `__slot__`" to learn more about it?

Changes look good. Well done.

illicitonion:

PR #172, opened 2026-06-21
NW | 26-SDC-Mar | TzeMing Ho | Sprint 2 | improve with precomputing

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

I have attempted with sorted and set

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -15,7 +15,7 @@ def find_longest_common_prefix(strings: List[str]):
             longest = common
     return longest
 # before changes, it was N^2, bacause it got a for loop inside a for loop
-# after changes, it is N + N. The first N for sorted, and another N for the loop.
+# after changes, it is NlogN + N. The NlogN is the worst case for sorted, and another N for the loop.
 
 def find_common_prefix(left: str, right: str) -> str:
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Code looks good.

Could you use Big O notation to quantify the improvement achieved by the new implementation? A brief comparison of the time complexity before and after the change would be helpful.

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
```diff
@@ -14,7 +14,8 @@ def find_longest_common_prefix(strings: List[str]):
         if len(common) > len(longest):
             longest = common
     return longest
-
+# before changes, it was N^2, bacause it got a for loop inside a for loop
+# after changes, it is N + N. The first N for sorted, and another N for the loop.
```
cjyuan: Could you lookup the complexity of `sorted()` in Python?

TzeMingHo: The worst case for sorted is NlogN, when the list is randomly ordered.

cjyuan: O(n log n) is the optimal average-case time complexity achievable by any comparison-based sorting algorithm.

An O(n) complexity approach is possible but that would involve writing a lot of code. 😄

illicitonion:

PR #177, opened 2026-06-24
NW | 26-SDC-Mar | Zabihollah Namazi | Sprint 2 | improve with caches

ZabihollahNamazi:

Hi! Could you please check my code?

I changed the Fibonacci and Making Change code. The old code was too slow and crashed with big numbers. I fixed it by using caching and loops so the code runs fast and passes all the big tests instantly.

Thank you!

### DIFF
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,4 +1,17 @@
 def fibonacci(n):
+    if n < 0:
+        raise ValueError("Fibonacci is not defined for negative integers.")
     if n <= 1:
         return n
-    return fibonacci(n - 1) + fibonacci(n - 2)
+        
+    # Track the last two numbers in the sequence
+    prev2 = 0  # fibonacci(0)
+    prev1 = 1  # fibonacci(1)
+    
+    # Calculate upward to n
+    for _ in range(2, n + 1):
+        current = prev1 + prev2
+        prev2 = prev1
+        prev1 = current
+        
+    return prev1
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,17 +1,13 @@
-def fibonacci(n):
-    if n < 0:
-        raise ValueError("Fibonacci is not defined for negative integers.")
-    if n <= 1:
-        return n
+fib_cache = {}
+def fibonacci(n: int) -> int:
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/fibonacci/fibonacci.py
```diff
@@ -1,4 +1,17 @@
...
+    prev2 = 0  # fibonacci(0)
+    prev1 = 1  # fibonacci(1)
+    
+    # Calculate upward to n
+    for _ in range(2, n + 1):
+        current = prev1 + prev2
+        prev2 = prev1
+        prev1 = current
+        
+    return prev1
```
cjyuan: Replacing the recursive approach with an iterative one defeats the purpose of this exercise, which is to improve the performance of a recursive function using caching. Could you update the code while keeping the original recursive implementation?

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,12 @@
...
+    
+    dp = [0] * (total + 1)
+    
+    dp[0] = 1
+    
+    for coin in coins:
+        for i in range(coin, total + 1):
+            dp[i] += dp[i - coin]
+            
+    return dp[total]
```
cjyuan: Good job in finding out a faster approach to find the solution.

Could you also include an implementation that uses cache to speed up the original recursive implementation?

Sprint-2/improve_with_caches/fibonacci/fibonacci.py
ZabihollahNamazi: i have updated the code 
thanks

Sprint-2/improve_with_caches/making_change/making_change.py
ZabihollahNamazi: i have updated the code 
thanks

Sprint-2/improve_with_caches/fibonacci/fibonacci.py
```diff
@@ -1,17 +1,13 @@
-def fibonacci(n):
-    if n < 0:
-        raise ValueError("Fibonacci is not defined for negative integers.")
-    if n <= 1:
-        return n
+fib_cache = {}
+def fibonacci(n: int) -> int:
+    if n == 0:
+        return 0
```
cjyuan, line 4: Note: We could also keep the cache as an attribute of the function.

cjyuan (commented):
Changes look good.

illicitonion:

PR #178, opened 2026-06-24
NW | 26-SDC-Mar | Zabihollah Namazi | Sprint 2 | improve with precomputing 

ZabihollahNamazi:

Hi! Could you please check my code?
The old code used slow loops and would freeze with big data. I fixed them by using sorting and sets so they now run instantly and easily handle millions of characters.All tests are passing perfectly. Thank you!

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -2,17 +2,15 @@ from typing import List
 
 
 def find_longest_common_prefix(strings: List[str]):
-    """
-    find_longest_common_prefix returns the longest string common at the start of any two strings in the passed list.
-
-    In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
-    """
+    if len(strings) < 2:
+        return ""
+    sorted_strings = sorted(strings)
     longest = ""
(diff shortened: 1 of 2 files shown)


cjyuan (commented):
Code looks good.

Could you use complexity to explain how the new implementation is better than the original implementation?

ZabihollahNamazi:
-Common Prefix:
    Old Way (Slow): To find the best matching pair, the code had to compare every single word with every other word one by one. If you had 10,000 words, it took millions of slow steps.

New Way (Fast): By sorting the list first (O(N log N . M)) , words that look alike are automatically grouped right next to each other. Now, the code only needs to look at neighbor pairs in one quick pass. It cuts out thousands of useless comparisons!

-Count Letters
    Old Way (Slow): For every uppercase letter found, the code had to scan the entire text again from start to finish to see if a lowercase version existed. This repeated scanning made it freeze up on large text (O(N^2)).
    New Way (Fast): The code reads the text exactly once from start to finish to sort letters into two Hash Sets. Because checking a Python set is instant (O(1)), the whole program runs in a single, fast linear pass (O(N)).

cjyuan:
Explanation is clear. Well done.

Note: You could also just mention the complexity of the original Common Prefix algorithm is O(M*N^2) without giving an example of 10,000 words.

illicitonion:

PR #181, opened 2026-06-26
London | 26-SDC-Mar | Craig D'Silva | Sprint 2 | LRU cache

craig-dsilva:
## Learners, PR Template

Self checklist

## Changelist

Implement LRU cache

### DIFF
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -1,0 +1,20 @@
+import datetime
+
+class LruCache:
+    def __init__(self, limit):
+        if limit < 1:
+            raise ValueError("Cache limit should be more than or equal to 1")
+        self.cache = []
+        self.limit = limit
+
+    def set(self, key, value):
+        while len(self.cache) >= self.limit:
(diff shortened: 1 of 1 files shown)


Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, limit):
+        if limit < 1:
+            raise ValueError("Cache limit should be more than or equal to 1")
+        self.cache = []
+        self.limit = limit
+
+    def set(self, key, value):
+        while len(self.cache) >= self.limit:
+            oldest = min(self.cache, key=lambda data: data["time"])
+            self.cache.remove(oldest)
```
cjyuan, line 13: Why use `while` and not `if`?

cjyuan (commented):
Using timestamps to track the LRU item has a couple of drawbacks:
- On a very fast machine, multiple entries may end up with the same timestamp
- Finding the LRU entry with `min(self.cache)` is an O(n) operation.

Could you implement a more efficient LruCache so that both `set()` and `get()` operations are O(1) operations?

PR #184, opened 2026-06-28
Glasgow | 26- SDC-Mar | Taras Mykytiuk | Sprint 2 | Improve_with_cache

TarasMykytiuk:
## Learners, PR Template

Self checklist

## Changelist

Improve with cache exercise compete.

### DIFF
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,10 +1,8 @@
-cache = {}
-
-def fibonacci(n):
+def fibonacci(n, cache = {}):
     if n <= 1:
         return n
     if n in cache:
         return cache[n]
     else:
-        cache[n] = fibonacci(n - 1) + fibonacci(n - 2)
+        cache[n] = fibonacci(n - 1, cache) + fibonacci(n - 2, cache)
         return cache[n]
--- a/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
+++ b/Sprint-2/improve_with_caches/fibonacci/fibonacci.py
@@ -1,4 +1,10 @@
+cache = {}
+
 def fibonacci(n):
     if n <= 1:
         return n
-    return fibonacci(n - 1) + fibonacci(n - 2)
+    if n in cache:
+        return cache[n]
+    else:
+        cache[n] = fibonacci(n - 1) + fibonacci(n - 2)
+        return cache[n]
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,7 +1,4 @@
 from typing import List
-cache = {}
-COINS = [200, 100, 50, 20, 10, 5, 2, 1]
-COINT_TYPES_NUM = len(COINS)
 
 def ways_to_make_change(total: int) -> int:
     """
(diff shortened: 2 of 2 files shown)


Sprint-2/improve_with_caches/fibonacci/fibonacci.py
```diff
@@ -1,4 +1,10 @@
+cache = {}
+
 def fibonacci(n):
```
cjyuan: Could you research alternative approaches that don't require keeping the cache in the global scope?

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,40 @@
...
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+def ways_to_make_change_helper(total: int, coin_index: int) -> int:
     """
     Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
     """
-    if total == 0 or len(coins) == 0:
+    if (total, coin_index) in cache:
+        return cache[(total, coin_index)]
```
cjyuan: Could we reuse this tuple in this function (instead of recreating new ones)?

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,40 @@
...
     Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
 
     For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
     """
-    return ways_to_make_change_helper(total, [200, 100, 50, 20, 10, 5, 2, 1])
+    return ways_to_make_change_helper(total, 0)
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+def ways_to_make_change_helper(total: int, coin_index: int) -> int:
```
cjyuan: Could also keep the coin list in the parameter so that the function can be reused on different coin list. 

The cost of evaluating `len(list)` is also negligible because it merely reads an attribute of the list.

Sprint-2/improve_with_caches/making_change/making_change.py
TarasMykytiuk: Done.

Sprint-2/improve_with_caches/making_change/making_change.py
TarasMykytiuk: Done.

Sprint-2/improve_with_caches/fibonacci/fibonacci.py
TarasMykytiuk: Thanks, it is interesting advice.

cjyuan: Your new approach works. I mainly wanted you to explore some alternatives.

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,40 +1,42 @@
...
 
     For instance, there are two ways to make a value of 3: with 3x 1 coins, or with 1x 1 coin and 1x 2 coin.
     """
-    return ways_to_make_change_helper(total, 0)
+    coins = [200, 100, 50, 20, 10, 5, 2, 1]
+    return ways_to_make_change_helper(total, 0, coins)
 
 
-def ways_to_make_change_helper(total: int, coin_index: int) -> int:
+def ways_to_make_change_helper(total: int, coin_index: int, coins: List[int], cache = {}) -> int:
```
cjyuan, line 13: Please note that, unlike JS, default value in Python is only created once on function definition instead of on every function call.

For `fibonacci()`, the cache is reusable across function calls, so setting the cache as a default parameter is acceptable.

But for `ways_to_make_change_helper()`, if `coins` is allowed to contain different coin values, then creating the cache as a default parameter would not be appropriate.

cjyuan (commented):
All good.

illicitonion:

PR #191, opened 2026-06-30
West Midlands | 26 March SDC | Iswat Bello | Sprint 2 | Implement a linked list in python

Iswanna:
## Learners, PR Template

Self checklist

## Changelist

In this PR, I implemented a **Doubly Linked List** to handle data storage and removal efficiently. A detailed breakdown of the implementation and trade-offs can be found in the `CHANGES-MADE.md` file included in this submission.

#### Key Changes
- **Node & LinkedList Classes**: Defined a `Node` class with bidirectional pointers (`next` and `previous`) and a `LinkedList` class to manage the `head` and `tail`.
- **Method Implementation**:
    - `push_head(value)`: Adds a node to the front in **O(1) time**.
    - `pop_tail()`: Removes the last node in **O(1) time**, including logic to handle single-node and empty lists.
    - `remove(node)`: A robust method that re-links neighboring nodes to remove a specific entry from anywhere in the list.
- **Pointer Management**: Ensured that all pointers (`head`, `tail`, `next`, and `previous`) are correctly updated during every operation to maintain list integrity.

#### Testing Done
I verified the code using `unittest` with the following scenarios:
- Standard `push_head` and `pop_tail` sequences.
- **Middle Removal**: Verified that removing a center node correctly re-wires the nodes ahead and behind it.
- **Empty List Handling**: Confirmed that `pop_tail` returns `None` rather than crashing when the list is empty.
- **Null Safety**: Confirmed that `remove(None)` is handled gracefully.

#### Learning Reflection  (Trade-offs)
I applied a **space-vs-time trade-off** by using extra memory (for the `previous` pointer) to ensure that adding to the head and removing from the tail stays at **O(1) speed** regardless of how many items are in the list.

### DIFF
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -1,0 +1,9 @@
+class Node:
+    """
+    Represents a single person in line.
+    Stores a value and links to the people ahead and behind.
+    """
+    def __init__(self, value):
+        self.value = value
+        self.next = None      # The person ahead
+        self.previous = None  # The person behind
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -34,4 +34,23 @@ class LinkedList:
         # Logic 4: Return the node so we can reference it later
         return new_node
 
-    
+    def pop_tail(self):
+        """Removes the last node and returns its value."""
+        # Safety check: if list is empty
+        if self.tail is None:
+            return None
+
+        # Logic 1: Store the value to return later
+        value_to_return = self.tail.value
+
+        # Logic 2: If there is only one person in line
+        if self.head == self.tail:
+            self.head = None
+            self.tail = None
+        # Logic 3: If there are more people
+        else:
+            self.tail = self.tail.previous  # Move tail back one
+            self.tail.next = None           # New tail lets go of the old tail
+
+        # Logic 4: Return the name/value
+        return value_to_return
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -7,3 +7,12 @@ class Node:
         self.value = value
         self.next = None      # The person ahead
         self.previous = None  # The person behind
+
+
+class LinkedList:
+    """
+    Manages the line by keeping track of the Head and the Tail.
+    """
+    def __init__(self):
+        self.head = None
+        self.tail = None
--- a/Sprint-2/implement_linked_list/linked_list.py
+++ b/Sprint-2/implement_linked_list/linked_list.py
@@ -54,3 +54,28 @@ class LinkedList:
 
         # Logic 4: Return the name/value
         return value_to_return
+
+    def remove(self, node):
+        """Removes a specific node from anywhere in the list."""
+        if node is None:
+            return
+
+        # Logic 3: Update Head/Tail pointers if the node is at the ends
+        if node == self.head:
+            self.head = node.next
+        
+        if node == self.tail:
+            self.tail = node.previous
+
+        # Logic 1 & 2: Re-link the neighbors
+        # If there's someone behind, they skip over 'node' to grab 'node.next'
+        if node.previous is not None:
+            node.previous.next = node.next
+            
+        # If there's someone ahead, they reach back to grab 'node.previous'
+        if node.next is not None:
+            node.next.previous = node.previous
+
+        # Optional: Clean up the removed node's own pointers
+        node.next = None
+        node.previous = None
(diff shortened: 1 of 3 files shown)


Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,81 @@
+class Node:
+    """
+    Represents a single person in line.
+    Stores a value and links to the people ahead and behind.
+    """
+    def __init__(self, value):
+        self.value = value
+        self.next = None      # The person ahead
+        self.previous = None  # The person behind
```
cjyuan: May I suggest exploring the use of `__slots__` to reduce memory usage?

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Logic 2: If the list is empty
+        if self.head is None:
+            self.head = new_node
+            self.tail = new_node
+        # Logic 3: If the list is not empty
+        else:
+            new_node.next = self.head      # New person grabs old head's hand
+            self.head.previous = new_node  # Old head reaches back to new person
+            self.head = new_node           # New person is now the head
```
cjyuan: - `else` already means "opposite", so you could probably drop the comment on line 28.

- The comments on line 30-32 actually confuses me because it refers to "person". Are they your own notes? If so, it might not be a good idea to leave them in "production" code.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,81 @@
...
+
+class LinkedList:
+    """
+    Manages the line by keeping track of the Head and the Tail.
+    """
+    def __init__(self):
+        self.head = None
+        self.tail = None
+
+    def push_head(self, value):
```
cjyuan: Could consider adding type hints to each method to clearly specify the types of its parameters and return value.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,81 @@
...
+        value_to_return = self.tail.value
+
+        # Logic 2: If there is only one person in line
+        if self.head == self.tail:
+            self.head = None
+            self.tail = None
+        # Logic 3: If there are more people
+        else:
+            self.tail = self.tail.previous  # Move tail back one
+            self.tail.next = None           # New tail lets go of the old tail
```
cjyuan: Why not just call `remove()`?

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,81 @@
...
+        return new_node
+
+    def pop_tail(self):
+        """Removes the last node and returns its value."""
+        # Safety check: if list is empty
+        if self.tail is None:
+            return None
+
+        # Logic 1: Store the value to return later
+        value_to_return = self.tail.value
```
cjyuan: The variable name is self-explanatory. The comment does not provide any more info.

Sprint-2/implement_linked_list/linked_list.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Logic 2: If there is only one person in line
+        if self.head == self.tail:
+            self.head = None
+            self.tail = None
+        # Logic 3: If there are more people
```
cjyuan: What do these "logic numbers" represent?

Sprint-2/implement_linked_list/linked_list.py
Iswanna: Thank you for the suggestion. I have implemented `__slots__` to reduce memory usage.

Sprint-2/implement_linked_list/linked_list.py
Iswanna: Thank you for the feedback.
Those metaphors were part of my initial notes to help me understand the concept. I've cleaned up the code to remove the non-technical language and the redundant logic labels.

Sprint-2/implement_linked_list/linked_list.py
Iswanna: Thank you for the feedback. I have updated the code to implement your suggestions.

Sprint-2/implement_linked_list/linked_list.py
Iswanna: Thank you for the suggestion. I’ve refactored the `pop_tail` method to call `self.remove(self.tail)`

Sprint-2/implement_linked_list/linked_list.py
Iswanna: I have removed all redundant comments. Thank you.

Sprint-2/implement_linked_list/linked_list.py
Iswanna: They represented the steps of the logic, but I have now removed them.

PR #193, opened 2026-06-30
West Midlands | 26 March SDC | Iswat Bello | Sprint 2 | Implement an LRU cache in python

Iswanna:
## Learners, PR Template

Self checklist

## Changelist

In this PR, I have implemented a standalone **LRU (Least Recently Used) Cache**. This task required building a purpose-built Doubly Linked List and Dictionary hybrid to ensure **O(1) worst-case time complexity** for both `get` and `set` operations.

A detailed explanation of the architecture and the logic behind the pointer management is available in the **`CHANGES_MADE.md`** file.

#### Key Features
- **Standalone Implementation**: Developed a custom `Node` and `LinkedList` class specifically for the requirements of an LRU Cache, separate from previous tasks.
- **Bidirectional Nodes**: Designed the `Node` class to store both **keys and values**. This enables the "reverse lookup" required to synchronize the dictionary when a node is evicted from the tail of the list.
- **O(1) Efficiency**: By combining a dictionary for instant lookups and a linked list for instant ordering, the cache remains performant regardless of its size.
- **Eviction Logic**: Automatically removes the least recently used item when the specified `limit` is reached.

#### Testing Done
Verified the implementation with `unittest` in `lru_cache_test.py`, including:
- Standard set/get sequences and limit enforcement.
- **MRU Refreshing**: Ensuring `get` operations move items to the front of the list.
- **Edge Case - Limit 1**: Confirmed correct behavior for the smallest possible cache.
- **Update Logic**: Verified that overwriting an existing key updates the value and refreshes its position without increasing cache size.
- **Null Safety**: Handled non-existent keys and null inputs gracefully.

#### Learning Reflection
This implementation demonstrates the **Space-vs-Time trade-off**. By using extra memory for a dictionary and a bidirectional linked list, I ensured that the speed of the system is not sacrificed as the data set grows.

### DIFF
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -21,4 +21,18 @@ class LinkedList:
         self.head = None
         self.tail = None
 
-   
+    def push_head(self, node):
+        """Adds an existing node to the front of the list."""
+        node.next = self.head
+        node.previous = None
+        
+        if self.head is not None:
+            self.head.previous = node
+        
+        self.head = node
+        
+        # If list was empty, this node is also the tail
+        if self.tail is None:
+            self.tail = node
+
+    
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -66,3 +66,23 @@ class LinkedList:
         # Clean up pointers of the removed node
         node.next = None
         node.previous = None
+
+
+class LruCache:
+    """
+    The LRU Cache: Dictionary + Linked List.
+    Provides O(1) time complexity for both get and set operations.
+    """
+    def __init__(self, limit):
+        # 1. Validation
+        if limit < 1:
+            raise ValueError("Cache limit must be at least 1.")
+            
+        # 2. Store the limit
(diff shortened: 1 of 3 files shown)


Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Step 1: Updated Node class.
+    Each node stores both key and value so we can find the 
+    dictionary entry when a node is evicted from the tail.
+    """
+    def __init__(self, key, value):
+        self.key = key
+        self.value = value
+        self.next = None
+        self.previous = None
```
cjyuan, line 11: To practice code reuse, you could import the linked list you implemented in the other exercise and store the key-value pairs as a tuple in each node.

Iswanna:
> Looks good.

Thank you for the feedback.

Sprint-2/implement_lru_cache/lru_cache.py
Iswanna: Thank you for the suggestion. 

By storing the key and value together as a tuple inside the node, I could use the Linked List I already built without having to write a new one for this file. Thanks for the suggestion.

PR #194, opened 2026-06-30
London | 26-SDC-Mar | beko | Sprint 2 | implement an LRU cache in python

Abubakar-Meigag:
Self checklist

Completed implementing an LRU cache in Python

### DIFF
--- a/Sprint-2/implement_lru_cache/lru_cache.py
+++ b/Sprint-2/implement_lru_cache/lru_cache.py
@@ -1,61 +1,73 @@
 from typing import Optional
 
 class _Node:
+    
     def __init__(self, key=None, value=None):
-        self.key   = key
+        self.key = key
         self.value = value
         self.prev: Optional["_Node"] = None
         self.next: Optional["_Node"] = None
-
-class LruCache:
-    def __init__(self, limit):
-        if limit <= 0:
-            raise ValueError("limit must be positive")
-
-        self.limit = limit
-        self.cache = {}
-
+        
+class DoublyLinkedList:
+    
+    def __init__(self):
         self.head = _Node()
         self.tail = _Node()
         self.head.next = self.tail
         self.tail.prev = self.head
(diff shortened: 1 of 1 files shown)


cjyuan (commented):
To better adhere to the **Single-Responsibility Principle** (SRP) from SOLID design principles, it's preferable to implement the "doubly linked list" and the "LRU Cache" as separate classes, with the linked list used inside LruCache to manage ordering.

Alternatively, `OrderedDict` can be used directly within `LruCache` to maintain order.

Abubakar-Meigag:
thanks @cjyuan for the feedback

i've refactored the implementation and and follow SRP by splitting the doubly linked list logic into its own DoublyLinkedList class

it handles all the node ordering (remove, add_to_front, remove_last), while LruCache focuses purely on the cache logic key lookups, limit enforcement, and eviction. 

the two classes are now independent and each has a single clear responsibility.

Sprint-2/implement_lru_cache/lru_cache.py
```diff
@@ -0,0 +1,73 @@
+from typing import Optional
+
+class _Node:
+    
+    def __init__(self, key=None, value=None):
```
cjyuan, line 5: Why not reuse your LinkedList class from the other exercise?

cjyuan (commented):
Changes look good.

Sprint-2/implement_lru_cache/lru_cache.py
Abubakar-Meigag: you're right i can reuse linked_list from the other exercise, I just chose to keep the linked list logic self-contained within the LRU cache because the node structure here is a bit different and a good exercise to have a different approach

cjyuan: You could store key-value pairs as a tuple in the node.

PR #196, opened 2026-06-30
London| 26 March SDC | Jamal Laqdiem | Sprint 1 | Refactor Functions

jamallaqdiem:
Self checklist

## Changelist

Refactored the scripts js and py to optimise the time and space complexity.

### DIFF
--- a/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
+++ b/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
@@ -1,19 +1,19 @@
 /**
  * Finds common items between two arrays.
  *
- * Time Complexity:O(1)
- * Space Complexity:O(n+m)
- * Optimal Time Complexity:O(1)
+ * Time Complexity:O(n*m)
+ * Space Complexity:O(u)
+ * Optimal Time Complexity:O(n+m)
  *
  * @param {Array} firstArray - First array to compare
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
 export const findCommonItems = (firstArray, secondArray) => {
-  const secondArr = new Set(secondArray);
+  const allowedItems = new Set(secondArray);
 
   const uniqueValues = [
-    ...new Set(firstArray.filter((item) => secondArr.has(item))),
+    ...new Set(firstArray.filter((item) => allowedItems.has(item))),
   ];
   return uniqueValues;
 };
--- a/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
+++ b/Sprint-1/JavaScript/findCommonItems/findCommonItems.js
@@ -1,14 +1,21 @@
 /**
  * Finds common items between two arrays.
  *
- * Time Complexity:
- * Space Complexity:
- * Optimal Time Complexity:
+ * Time Complexity:O(1)
+ * Space Complexity:O(n+m)
+ * Optimal Time Complexity:O(1)
  *
  * @param {Array} firstArray - First array to compare
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
-export const findCommonItems = (firstArray, secondArray) => [
-  ...new Set(firstArray.filter((item) => secondArray.includes(item))),
-];
+export const findCommonItems = (firstArray, secondArray) => {
+  const secondArr = new Set(secondArray);
+
+  const uniqueValues = [
+    ...new Set(firstArray.filter((item) => secondArr.has(item))),
+  ];
+  return uniqueValues;
+};
+
+console.log(findCommonItems([2, 5, 7], [5, 6, 7]));
--- a/Sprint-1/JavaScript/hasPairWithSum/hasPairWithSum.js
+++ b/Sprint-1/JavaScript/hasPairWithSum/hasPairWithSum.js
@@ -1,21 +1,26 @@
 /**
  * Find if there is a pair of numbers that sum to a given target value.
  *
- * Time Complexity:
- * Space Complexity:
- * Optimal Time Complexity:
+ * Time Complexity:O(n)
+ * Space Complexity:O(n)
+ * Optimal Time Complexity:O(n)
  *
  * @param {Array<number>} numbers - Array of numbers to search through
  * @param {number} target - Target sum to find
  * @returns {boolean} True if pair exists, false otherwise
  */
 export function hasPairWithSum(numbers, target) {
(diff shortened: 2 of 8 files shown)


Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -1,14 +1,21 @@
...
  *
  * @param {Array} firstArray - First array to compare
  * @param {Array} secondArray - Second array to compare
  * @returns {Array} Array containing unique common items
  */
-export const findCommonItems = (firstArray, secondArray) => [
-  ...new Set(firstArray.filter((item) => secondArray.includes(item))),
-];
+export const findCommonItems = (firstArray, secondArray) => {
+  const secondArr = new Set(secondArray);
```
cjyuan: An `Arr` suffix tends to suggest the variable holds an array.

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
```diff
@@ -1,14 +1,21 @@
 /**
  * Finds common items between two arrays.
  *
- * Time Complexity:
- * Space Complexity:
- * Optimal Time Complexity:
+ * Time Complexity:O(1)
+ * Space Complexity:O(n+m)
+ * Optimal Time Complexity:O(1)
```
cjyuan: Could you explain how you derived the time complexity of $O(1)$ and the space complexity of $O(n+m)$?

Sprint-1/JavaScript/hasPairWithSum/hasPairWithSum.js
```diff
@@ -1,21 +1,26 @@
...
-  for (let i = 0; i < numbers.length; i++) {
-    for (let j = i + 1; j < numbers.length; j++) {
-      if (numbers[i] + numbers[j] === target) {
-        return true;
-      }
+  const seen = new Set();
+
+  for (let num of numbers) {
+    const complement = target - num;
+    console.log(complement);
```
cjyuan: Debugging code?

cjyuan (commented):
Code is correct.

- Could you also include the time complexity of the original implementation?

    Note: I think "Time Complexity" in the files refers to the time complexity of the original implementation, and "Optimal Time Complexity" refers to the time complexity of the refactored version.

Sprint-1/JavaScript/hasPairWithSum/hasPairWithSum.js
jamallaqdiem: I used for a specific purpose at the start of the refactoring, however I should deleted.

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
jamallaqdiem: Thanks for taking time to review, the arr is misleading to an array, instead the variable now hold a set, I just changed.

Sprint-1/JavaScript/findCommonItems/findCommonItems.js
jamallaqdiem: Thanks for reviewing, is the the Optimal time complexity that should be O(n+m),  we have to spend O(m) time to build the set and O(n) time to filter firstArray.
the space Com should be O(u) to store the matching items.

cjyuan: What is $u$? Normally we express all complexity based on the size of the input, which are $n$ and $m$.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -9,8 +9,8 @@ def find_common_items(
     """
     Find common items between two arrays.
 
-    Time Complexity: O(n+m)
-    Space Complexity:O(n+m)
+    Time Complexity: O(n^3)
+    Space Complexity:O(u)
```
cjyuan: Could you explain why the complexity of the original code is $O(n^3)$? And what is $u$?

jamallaqdiem: The  O(n³) comes because we have 3 loops the main one, the nested loop, then  another for i in common_items each have O(n).
we should use O(m(n,m)) , n and m are the length of the two sequences.

Sprint-1/Python/find_common_items/find_common_items.py
```diff
@@ -9,13 +9,16 @@ def find_common_items(
     """
     Find common items between two arrays.
 
-    Time Complexity:
-    Space Complexity:
-    Optimal time complexity:
+    Time Complexity: O(n^3)
+    Space Complexity:O(min(n,m))
+    Optimal time complexity: O(n+m)
```
cjyuan, line 14: - Why use $m$ in the complexity of your algorithm but not in the complexity of the original algorithm?

- How does the algorithm use $O(min(n, m))$ space?

jamallaqdiem: You are completely right about the inconsistency between old and new  Time complex:
1. let's define n as len(first_sequence) and m as len(second_sequence), if the input sizes differ, m should be used in both.
Original Algorithm: as explained earlier the outer loop runs n times, the inner loop runs m times, and the not in list takes up to O(n) time. Therefore, time complexity of the original code is O(n^2 \cdot m). If we assume n = m, it simplifies to O(n^3).
New Algorithm: Converting the sequences to sets takes O(n) + O(m) time. The Python set intersection '&' average-case time complexity is O(\min(n, m)) as documented in this link :(https://wiki.python.org/moin/TimeComplexity). the overall optimal time complexity of O(n + m).

2. Space Complexity Proof for O(\min(n, m)):
the original algorithm allocates memory for the common_items list, an item is only appended to common_items  if it exists in both first_sequence and second_sequence
Mathematically, the intersection of two sets cannot contain more elements than the smaller of the two sets

cjyuan: How much space needed by `first_set` and `second_set`?

jamallaqdiem: Now I got it, n the original script first_set and  second_set do not exist in that implementation, so they take up no additional space.
because the input is already in memory, in this function we did not allocated any extra space, just pointing where they live in memory.

cjyuan: It wasn't clear whether the space complexity analysis is for the original or the new implementation.
I was expecting it for the new implementation but you thought it was for the original code.

In that case, all good.

PR #198, opened 2026-07-01
West Midlands | 26-SDC-Mar | Chioma Okeke | Sprint 2 | Improve code with caches

JanefrancessC:
## Learners, PR Template

Self checklist

## Changelist
Completed improve code with caches exercise

### DIFF
--- a/Sprint-2/improve_with_caches/making_change/making_change.py
+++ b/Sprint-2/improve_with_caches/making_change/making_change.py
@@ -1,3 +1,5 @@
+COINS = (200, 100, 50, 20, 10, 5, 2, 1)
+
 def ways_to_make_change(total: int) -> int:
     """
     Given access to coins with the values 1, 2, 5, 10, 20, 50, 100, 200, returns a count of all of the ways to make the passed total value.
@@ -11,7 +13,6 @@ def ways_to_make_change_helper(total: int, coin_index: int, cache={}) -> int:
     """
     Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
     """
-    coins = (200, 100, 50, 20, 10, 5, 2, 1)
 
     key = (total, coin_index)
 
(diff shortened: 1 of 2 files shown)


Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,35 @@
...
+    return ways_to_make_change_helper(total, 0)
 
 
-def ways_to_make_change_helper(total: int, coins: List[int]) -> int:
+def ways_to_make_change_helper(total: int, coin_index: int, cache={}) -> int:
     """
     Helper function for ways_to_make_change to avoid exposing the coins parameter to callers.
     """
-    if total == 0 or len(coins) == 0:
+    coins = (200, 100, 50, 20, 10, 5, 2, 1)
```
cjyuan: Declaring `coin` as a local variable means a different tuple (with identical items) is created in every recursive function call.

Why not pass the coin list through the parameter so that the function can be reused for different coin list?

However, if the function is expected to accept different coin lists, it would not be appropriate to declare the cache as a parameter with a default value.

Sprint-2/improve_with_caches/making_change/making_change.py
```diff
@@ -1,32 +1,35 @@
...
-            else:
-                intermediate = ways_to_make_change_helper(total - total_from_coins, coins=coins[coin_index+1:])
-                ways += intermediate
-            count_of_coin += 1
+    coin = coins[coin_index]
+
+    if coin > total:
+        ways = ways_to_make_change_helper(total, coin_index + 1)
+    else:
+        ways = (ways_to_make_change_helper(total - coin, coin_index) + ways_to_make_change_helper(total, coin_index + 1))
```
cjyuan: It seems the code on lines 30-32 could be shorten.

JanefrancessC: The if/else structure, once flattened, breaks the flow for large inputs like the `(ways_to_make_change(9176))`

Sprint-2/improve_with_caches/making_change/making_change.py
JanefrancessC: I have refactored and made `COINS` a global variable

Sprint-2/improve_with_caches/making_change/making_change.py
cjyuan: One of the function calls will always happen. That would suggest it does not need to reside in an if-else block.

Sprint-2/improve_with_caches/making_change/making_change.py
cjyuan: Why not pass the coin list through the parameter?

PR #203, opened 2026-07-01
West Midlands | 26 March SDC | Iswat Bello | Sprint 2 | Improve code with precomputing

Iswanna:
## Learners, PR Template

Self checklist

## Changelist

In this PR, I have implemented performance optimisations for the `common_prefix` and `count_letters` tasks by applying pre-computing strategies. These changes successfully resolve $O(N^2)$ bottlenecks, allowing the functions to process millions of items in milliseconds.

I have provided a detailed technical breakdown of the complexity changes and architectural decisions in the **`CHANGES_MADE.md`** file.

#### Key Changes

**1. Longest Common Prefix (Task 5)**
- **Optimization**: Introduced a **pre-sorting** step using `strings.sort()`.
- **Logic**: By sorting the list alphabetically, I ensured that strings with common prefixes are neighbours. This allowed me to replace a nested loop with a single-pass comparison of adjacent strings.
- **Performance**: Transitioned from **$O(N^2)$** to **$O(N \log N)$**.

**2. Count Letters (Task 6)**
- **Optimization**: Introduced a **pre-computed Set** (`set(s)`) to handle character lookups.
- **Logic**: By converting the string into a set upfront, I replaced the $O(N)$ string scanning "hidden loop" (`not in s`) with a constant-time **$O(1)$** lookup.
- **Performance**: Transitioned from **$O(N^2)$** to **$O(N)$**.

#### Legacy Preservation
- I have made a conscious effort to retain legacy variable names (`longest`, `only_upper`, `string`) and re-use original helper functions (`find_common_prefix`, `is_upper_case`).

#### Testing Done
- Verified that all unit tests pass.
- **Scalability Test**: Confirmed that `test_really_long_list` (1,000,000 strings) and `test_long_string` (10,000,000 characters) now finish nearly instantly, whereas the original code would hang indefinitely.

#### Learning Reflection
These tasks highlighted the **Space-vs-Time trade-off**. By using a small amount of extra memory to store a sorted list or a set, we drastically reduced the CPU time required, which is essential for building scalable software.

### DIFF
--- a/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
+++ b/Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
@@ -1,18 +1,29 @@
 from typing import List
 
-
 def find_longest_common_prefix(strings: List[str]):
     """
-    find_longest_common_prefix returns the longest string common at the start of any two strings in the passed list.
-
-    In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
+    Optimized version using pre-sorting while keeping legacy names and helpers.
     """
+    # Edge case handling
+    if len(strings) < 2:
+        return ""
+
+    # Pre-compute (Sort) - This is the optimization!
+    strings.sort()
+
     longest = ""
-    for string_index, string in enumerate(strings):
-        for other_string in strings[string_index+1:]:
-            common = find_common_prefix(string, other_string)
-            if len(common) > len(longest):
-                longest = common
+
+    # Single loop replacing the nested loop
+    for string_index in range(len(strings) - 1):
+       
+        string = strings[string_index]
+        other_string = strings[string_index + 1]
+
+        common = find_common_prefix(string, other_string)
+
+        if len(common) > len(longest):
+            longest = common
+            
     return longest
 
 
--- a/Sprint-2/improve_with_precomputing/CHANGES_MADE.md
+++ b/Sprint-2/improve_with_precomputing/CHANGES_MADE.md
@@ -1,0 +1,46 @@
+# Changes Made: Optimization via Pre-computing
+
+## Overview
+These changes apply the strategy of **Pre-computing**: doing work upfront (like sorting or indexing) to make the primary search or calculation significantly faster. This allowed both algorithms to scale from small examples to datasets containing millions of items.
+
+---
+
+## 1. Longest Common Prefix Optimization (`common_prefix.py`)
+
+### The Problem
+The original implementation used a nested loop to compare every string in a list against every other string. 
+- **Original Complexity**: $O(N^2)$.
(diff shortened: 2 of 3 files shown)


Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
```diff
@@ -1,18 +1,29 @@
...
-
-    In the event that an empty list, a list containing one string, or a list of strings with no common prefixes is passed, the empty string will be returned.
+    Optimized version using pre-sorting while keeping legacy names and helpers.
     """
+    # Edge case handling
+    if len(strings) < 2:
+        return ""
+
+    # Pre-compute (Sort) - This is the optimization!
+    strings.sort()
```
nedssoft, line 12: This sorting step is the heart of your optimisation and it's spot-on. One subtle thing worth noticing, though: `strings.sort()` sorts the list *in place* — it reorders the very list the caller handed you. The original function only ever *read* the list, so whoever called it could rely on their list staying in its original order afterwards; with `.sort()`, it silently comes back rearranged.

For a test that just checks the return value it makes no difference — but imagine a caller who does something else with their list right after calling you. Python gives you a close cousin of `.sort()` that returns a *new* sorted list instead of changing the original in place — do you know which one? What would swapping to it cost you here, and what would it protect the caller from?

Your result is correct — this is just about being a considerate houseguest with data you don't own.

Iswanna:
> Iswat — two really nicely-reasoned optimisations here, and again your `CHANGES_MADE.md` does a lovely job of explaining the _why_, not just the _what_. The standout for me is the common-prefix insight: spotting that once the list is sorted, the strings sharing the longest prefix must be neighbours, so a single adjacent-pair pass replaces the whole nested loop. That's a genuinely clever result and it's mathematically sound. 👏 And the `set(s)` trick in `count_letters` — turning an O(N) `not in s` scan into an O(1) lookup while leaving the rest of the logic untouched — is exactly the right instinct.
> 
> One optional question inline — about a side-effect, not correctness (all your tests pass and both results are right). You've clearly met the task, so I'm marking this **Complete**. Really strong work across this sprint. 🎉

Thank you so much for the feedback. I will have a look at the inline question right away.

Sprint-2/improve_with_precomputing/common_prefix/common_prefix.py
Iswanna: Thanks for the feedback. I see now that `strings.sort()` reorders the caller's actual data, which could cause bugs for them later.
To be a better houseguest, I could use the `sorted()` function instead. The 'cost' would be Memory, because `sorted()` creates a new copy of the list instead of changing the original one. However, the 'protection' is that the actual list stays in its original order.
I'll remember to use `sorted()` in the future when I don't want to change the original data

